PR #103, opened 2022-03-03
feat(PPDSC-2032): audio player time display

baburay23:
PPDSC-2032

**What**

1. Background - created a CalculateTime function which converts seconds into the desired format. This function is also used by the user to display either the currentTime or the length
`<TimeDisplay format={({length}) => calculateTime(length)} />`
I also created FormatFunction() which the user can use to switch round the time stamps if needed
eg:   `<TimeDisplay format={({currentTime, length}) => formatFunction(length, currentTime)}/>`

i have also created scenarios once the slider is merged I will add it into the scenarios for the time display

3. 
4. What did you do
5. What does the reviewers should expect

**I have done:**

**I have tested manually:**

**Before**

**After:**

**Who should review this PR**

**How to test**


src/audio-player-composable/components/time-display/time-display.tsx
```diff
@@ -0,0 +1,32 @@
...
+`;
+
+export const ThemelessTimeDisplay = ({
+  overrides = {},
+  format,
+}: StyledLabelProps) => {
+  const {getTimeDisplayProps} = useAudioPlayerContext();
+  const {defaultFormat, currentTime, length} = getTimeDisplayProps!();
+  return (
+    <TimeDisplayContainer>
```
mutebg, line 21: Why do we use Label here? IMO that is not the semantic HTML tag, Label is used to describe input element and in our case do not have a such. Its better to use TextBlock as "span"

mutebg, line 21: Why do we use Label here? IMO that is not the semantic HTML tag, Label is used to describe input element and in our case do not have a such. Its better to use TextBlock as "span" and we do that we can remove `TimeDisplayContainer` and pass stylePreset and typographyPreset directly on it.

src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -100,9 +101,16 @@ export const AudioPlayerComposable = ({
     };
   };
 
+  const getTimeDisplayProps = () => ({
+    defaultFormat: formatFunction(trackPositionArr![0], duration!),
+    currentTime: trackPositionArr[0],
```
mutebg, line 106: @Xin00163 in your ticket this variable is named trackPosition but here is named: currentTime, maybe we should keep naming consistent in both functions.

src/audio-player-composable/components/time-display/time-display.tsx
Vanals: Yeah. agree!

src/audio-player-composable/audio-player-composable.tsx
Vanals: I personally would prefer keeping currentTime for the "label", more clear on what it is for, in the getTimeDisplay, there is not literally any track position

src/audio-player-composable/__tests__/utils.test.tsx
```diff
@@ -34,6 +35,15 @@ test('formatDurationTime', () => {
...
+  expect(formatFunction(1000.0, 5.0)).toEqual('16:40/00:05 ');
+  expect(formatFunction(23.0, 109000.0)).toEqual('00:23/30:16:40 ');
+  expect(formatFunction(120, 109000.0)).toEqual('02:00/30:16:40 ');
+  expect(formatFunction(4100, 109000.0)).toEqual('01:08:20/30:16:40 ');
+  expect(formatFunction(109000.0, 109000.0)).toEqual('30:16:40/30:16:40 ');
+});
+
 // describe('formatTrackData', () => {
 //   test('when fully buffered', () => {
 //     expect(
```
Vanals, line 49: is this a left over?

src/audio-player-composable/audio-player-composable.tsx
mutebg: @Vanals I agree, that's why my suggestion is updating trackPosition to currentTime in Xin's ticket. In reality, track position is calculated based on currentTime and duration.

src/audio-player-composable/components/time-display/time-display.tsx
baburay23: ah sorry i have now changed it to TextBlock however when i use `TextBlock as "span" ` without seekBa
Screenshot 2022-03-11 at 15 37 42
r

src/audio-player-composable/components/time-display/style-presets.ts
```diff
@@ -0,0 +1,14 @@
...
+  currentTimeDisplay: {
+    base: {
+      color: '{{colors.inkContrast}}',
+    },
+  },
+  container: {
+    base: {
+      backgroundColor: '{{colors.transparent}}',
+    },
+  },
```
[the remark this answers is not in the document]
baburay23: i did this because in [figma ](https://www.figma.com/file/FSbCQa6SzVR3K48ZWLeD77/branch/KVZQUwYoA0WqA289Y7LGHu/%F0%9F%9F%A2-NK-Web-Components?node-id=2151%3A0)

src/audio-player-composable/components/time-display/time-display.tsx
mutebg: that's an interesting case, in general, it does not matter a lot since it's very likely this to be used together with GridLayout or Stack component in order to build the whole audio player. 
Maybe you can change the display: to inline-block using styled TextBlock

src/audio-player-composable/components/time-display/time-display.tsx
```diff
@@ -1,28 +1,29 @@
...
 import stylePresets from './style-presets';
-import {Label} from '../../../label';
+
 import {StyledLabelProps} from './types';
+import {TextBlock} from '../../../text-block';
 
-const TimeDisplayContainer = styled.div<StyledLabelProps>`
+const StyledTextBlock = styled(TextBlock)<StyledLabelProps>`
   ${getStylePreset(`container`, '')}
+  ${getTypographyPreset(`timeDisplay.container`, '')};
```
mutebg, line 13: I think these paths are wrong, since we don't have container anymore:
```
${getStylePreset(`timeDisplay`, '')};
 ${getTypographyPreset(`timeDisplay`, '')};
 ```

File: src/audio-player-composable/components/time-display/style-presets.ts [hunk 6]
```diff
@@ -1,11 +1,6 @@
 import {StylePreset} from '../../../theme/types';
 
 export default {
-  currentTimeDisplay: {
-    base: {
-      color: '{{colors.inkContrast}}',
-    },
-  },
   container: {
```
mutebg, line 4: should this be the opposite?  you need to remove the container and leave the currentTimeDisplay?

src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -102,16 +102,23 @@ export const AudioPlayerComposable = ({
     };
   };
 
+  const getSeekBarProps = () => ({
+    duration,
+    trackPosition,
```
mutebg, line 107: I know this is not part of your ticket but if @Vanals and @Xin00163 agree with that let's rename trackPosition to currentTime so its consistent between both getter functions

src/audio-player-composable/components/time-display/defaults.ts
```diff
@@ -0,0 +1,10 @@
+export default {
+  timeDisplay: {
+    container: {
+      stylePreset: 'currentTimeDisplay',
+    },
+    currentTimeDisplay: {
+      stylePreset: 'currentTimeDisplay',
+    },
+  },
```
mutebg, line 9: This default file should be updated, if we have only TextBlock probably should like like
```
timeDisplay: {
  stylePreset: 'currentTimeDisplay',
}
```
and why there is not default typography preset?

src/audio-player-composable/context.tsx
```diff
@@ -1,5 +1,6 @@
 import {ReactJSXElement} from '@emotion/react/types/jsx-namespace';
 import React, {createContext, SyntheticEvent, useContext} from 'react';
+import {FormatFn} from './components/time-display/types';
```
mutebg, line 4: Since this is the kind of the last ticket before release, can we update useAudioPlayerContext to look like the menu context hook, so that customers have a warning if they try to use audio components outside audio context.
```
export const useAudioPlayerContext = () => {
  const context = useContext(AudioPlayerContext);

  /* istanbul ignore if */
  if (
    process.env.NODE_ENV !== 'production' &&
    Object.keys(context).length === 0
  ) {
    // eslint-disable-next-line no-console
    console.error(
      'You are using a component which needs to be a child of <AudioPlayerComposable />',
    );
  }

  return context;
};

```

src/audio-player-composable/components/time-display/style-presets.ts
baburay23: sorry for the confusion i now only have one style preset
```
 timeDisplay: {
    base: {
      backgroundColor: '{{colors.transparent}}',
    },
  },
```
and i have added typographyPreset to the defaults

src/audio-player-composable/components/time-display/time-display.tsx
baburay23: cool i have changed it back to span and added in `display: to inline-block `

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
```diff
@@ -64,6 +54,39 @@ const myCustomThemeTimeDisplay = createTheme({
...
+export const AudioPlayer = () => (
+  <>
+    <StorybookSubHeading>Audio player inline</StorybookSubHeading>
+    <AudioPlayerComposable
+      src="https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3"
+      ariaLandmark="audio player inline"
+    >
+      <GridLayout
+        columns="auto auto 1fr auto"
+        columnGap="16px"
```
mutebg, line 72: can we use a token here, should be `space040`

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
```diff
@@ -64,6 +54,39 @@ const myCustomThemeTimeDisplay = createTheme({
...
+        alignItems="center"
+      >
+        <PlayPauseButton
+          size={ButtonSize.Small}
+          onClick={() => {
+            console.log('customer click function');
+          }}
+        />
```
mutebg, line 80: Just to simplify it, can we remove the onClick event

src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -104,13 +104,13 @@ export const AudioPlayerComposable = ({
 
   const getSeekBarProps = () => ({
     duration,
-    trackPosition,
+    currentTime,
```
mutebg, line 107: if you rename trackPosition to currentTime my suggestion is to rename setTrackPosition and trackPositionRef.

src/audio-player-composable/components/time-display/time-display.tsx
```diff
@@ -4,13 +4,13 @@ import {useAudioPlayerContext} from '../../context';
...
 import defaults from './defaults';
 import stylePresets from './style-presets';
-
 import {StyledLabelProps} from './types';
 import {TextBlock} from '../../../text-block';
 
 const StyledTextBlock = styled(TextBlock)<StyledLabelProps>`
-  ${getStylePreset(`container`, '')}
-  ${getTypographyPreset(`timeDisplay.container`, '')};
+  display: inline-block;
```
mutebg, line 11: Maybe I made a mistake with that suggestion, 
Probably we should keep the component as a div and remove the display: inline-block.
as far I am aware all components we produce have a width of 100%.
Lets ask Mike about this

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
```diff
@@ -159,8 +187,14 @@ export const AudioPlayerWithTimeDisplay = () => (
...
       />
+      <Block spaceStack="space030" />
       <SeekBar />
-      <TimeDisplay format={({currentTime}) => calculateTime(currentTime)} />
+      <Block spaceStack="space030" />
+      <GridLayout columns="1fr 1fr 0fr 0fr 0fr">
+        <GridLayoutItem justifySelf="start">
+          <TimeDisplay format={({currentTime}) => calculateTime(currentTime)} />
+        </GridLayoutItem>
+      </GridLayout>
```
mutebg, line 197: I am not sure what is this GridLayout doing, can you elaborate please

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
```diff
@@ -174,44 +208,59 @@ export const AudioPlayerWithTimeDisplay = () => (
           console.log('customer click function');
         }}
       />
-      <TimeDisplay format={({duration}) => calculateTime(duration)} />
+      <Block spaceStack="space030" />
+      <SeekBar />
+      <Block spaceStack="space030" />
+      <GridLayout columns="repeat(1, 1fr)">
```
mutebg, line 214: same here

mutebg (commented):
Shall we add prefix name to all all audio-player components,
for example all form related are FormInputSelect, FormInputCheckbox, FormInputTextField.
SeekBar -> AudioPlayerSeekBar
PlayPauseButton -> AudioPlayerPlayPauseButton and so on
@Vanals @Xin00163 @pdimova  what you think?

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
baburay23: i have removed from all stories just left it on the playPauseButton story

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
baburay23: hey, I was trying to display the time at the start for this one. I see that with using justifySelf="start"  I was able to apply
`<TimeDisplay format={({currentTime}) => calculateTime(currentTime)} />` at the start and design wanted it to be in a grid. Is there a better way to do this? After what you showed me yesterday i tried to use it without `<GridLayoutItem/>` but i need to just wrap the <TimeDisplay/> component which is why i used it. 
I noticed that if i dont use `<GridLayoutItem/>` the colour jumps out
Screenshot 2022-03-15 at 11 13 01

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
baburay23: i was trying to do the same thing as above just place the <TimeDisplay/> at the start or end of the seek bar

src/audio-player-composable/components/time-display/time-display.tsx
baburay23: I spoke to Mike  and he said since its kinda unlikely they would use the time stamp on its own it does not really matter. but he gave his opinion of using span

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
baburay23: just fyi for anyone else  Since we are using span i have added in display: block; into the styled <TextBlock/> so the component should be at 100% width if it is not wrapped in a grid/stack etc (have spoken to Mike about this as well)

src/audio-player-composable/components/time-display/defaults.ts
```diff
@@ -0,0 +1,6 @@
+export default {
+  timeDisplay: {
```
mutebg, line 2: I am afraid that we need to add prefix to all defaults 
timeDisplay -> audioPlayerTimeDisplay
seekBar -> audioPlayerSeekBar

the reason is that at the moment we add volume control it will clash with our current volumeControl defaults

mutebg (changes requested):

src/audio-player-composable/components/seek-bar/seek-bar.tsx
```diff
@@ -20,9 +20,9 @@ const ThemelessSeekBar: React.FC<SeekBarProps> = ({overrides = {}}) => {
 
   const theme = useTheme();
   const {
-    seekBar: {slider: seekBarSliderDefaults},
+    audioPlayerSeekBar: {slider: seekBarSliderDefaults},
   } = theme.componentDefaults;
-
+  console.log(seekBarSliderDefaults);
```
mutebg, line 25: remove the console.log

src/audio-player-composable/components/time-display/time-display.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {withOwnTheme} from '../../../utils/with-own-theme';
+import defaults from './defaults';
+import stylePresets from './style-presets';
+import {StyledLabelProps} from './types';
+import {TextBlock} from '../../../text-block';
+
+const StyledTextBlock = styled(TextBlock)<StyledLabelProps>`
+  display: block;
+  ${getStylePreset(`timeDisplay`, '')}
+  ${getTypographyPreset(`timeDisplay`, '')};
```
mutebg, line 13: you need to update these 2 paths

mutebg (approved):

mutebg (dismissed):

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
```diff
@@ -1,10 +1,16 @@
 /* eslint-disable no-console */
 import * as React from 'react';
-import {PlayPauseButton} from '../components/play-pause-button/play-pause-button';
-import {SeekBar} from '../components/seek-bar/seek-bar';
+import {AudioPlayerPlayPauseButton} from '../components/play-pause-button/play-pause-button';
+import {AudioPlayerTimeDisplay} from '../components/time-display/time-display';
```
Vanals, line 4: if we have an index.js we should be able to do 
```'../components/time-display```
 without repetition, same for the other imports

File: .next/package.json [hunk 20]
```diff
@@ -0,0 +1 @@
+{"type": "commonjs"}
```
mutebg, line 1: can you remove this file, apparently it appears sometime

mutebg (approved):

src/audio-player-composable/context.tsx
```diff
@@ -47,4 +54,19 @@ export const AudioPlayerContext = createContext<
...
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    Object.keys(context).length === 0
+  ) {
+    // eslint-disable-next-line no-console
+    console.error(
+      'You are using a component which needs to be a child of <AudioPlayerComposable />',
+    );
+  }
+
```
Vanals, line 70: How will this work? in which scenarios this error will trigger 🤔 ?
This line `Object.keys(context).length === 0` seems to be key but I don't get it

Vanals (approved):

src/audio-player-composable/context.tsx
mutebg: this error will be shown when you run code in dev mode and you are trying to use a component out of the context

PR #108, opened 2022-03-08
feat(PPDSC-2031): audio player seekbar

Xin00163:
PPDSC-2031

- Separate the seekBar to an independent component
- Add stories for default and overrides

I've not implemented the layout as it makes more sense to apply grid once we have the controls.

**I have done:**

**I have tested manually:**

**Before**

**After:**

**Who should review this PR**

**How to test**

### DIFF
--- a/src/audio-player-composable/audio-player-composable.tsx
+++ b/src/audio-player-composable/audio-player-composable.tsx
@@ -17,13 +17,14 @@ export const AudioPlayerComposable = ({
   ariaLandmark,
 }: AudioPlayerComposableProps) => {
   const trackPositionRef = useRef(0);
+
   const audioRef = useRef<HTMLAudioElement>(null);
   const showLoaderTimeoutRef: MutableRefObject<number> = useRef(0);
 
   const [playing, setPlayState] = useState(autoPlay);
   const [loading, setLoading] = useState(true);
   const [duration, setDuration] = useState(0);
-  const [trackPositionArr, setTrackPosition] = useState([0]);
+  const [trackPosition, setTrackPosition] = useState(0);
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [displayDuration, setDisplayDuration] = useState(0);
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
@@ -32,11 +33,11 @@ export const AudioPlayerComposable = ({
   const [buffered, setBuffered] = useState<TimeRanges>();
 
   useEffect(() => {
-    [trackPositionRef.current] = trackPositionArr;
+    trackPositionRef.current = trackPosition;
   });
 
   useEffect(() => {
-    setTrackPosition([0]);
+    setTrackPosition(0);
     setDisplayDuration(0);
   }, [src]);
 
@@ -102,8 +103,8 @@ export const AudioPlayerComposable = ({
 
   const getSeekBarProps = () => ({
     duration,
-    trackPositionArr,
-    onChangeSlider,
+    trackPosition,
+    onChange: onChangeSlider,
     buffered,
   });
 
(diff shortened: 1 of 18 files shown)


src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -100,26 +100,23 @@ export const AudioPlayerComposable = ({
     };
   };
 
+  const getSeekBarProps = () => ({
+    duration,
+    trackPositionArr,
+    onChangeSlider,
```
mutebg, line 106: if we can rename that so its user friendly
```suggestion
    onChange: onChangeSlider,
```

src/audio-player-composable/components/seek-bar/seek-bar.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+const ThemelessSeekBar: React.FC<SeekBarProps> = ({overrides = {}}) => {
+  const {slider: seekBarSliderOverrides} = overrides;
+
+  const theme = useTheme();
+  const {
+    seekBar: {slider: seekBarSliderDefaults},
+  } = theme.componentDefaults;
+
+  const {getSeekBarProps} = useContext(AudioPlayerContext);
```
mutebg, line 26: can we create a separate hook that returns the context?

in the context.tsx file
```
export const useAudioPlayerContext = useContext(AudioPlayerContext);
```

```suggestion
  const {getSeekBarProps} = useAudioPlayerContext();
```

it's no big deal but that makes things a bit neater 

ps, update the PlayPause component to use the hook

src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -100,26 +100,23 @@ export const AudioPlayerComposable = ({
     };
   };
 
+  const getSeekBarProps = () => ({
+    duration,
+    trackPositionArr,
```
mutebg, line 105:
```suggestion
    trackPosition,
```

src/audio-player-composable/utils.ts
```diff
@@ -12,44 +12,41 @@ export const formatDuration = (time: number, maxTime: number = time) => {
   }
   return formatTrackTime(time, maxTime);
 };
+export const formatTrackData = (
+  trackColor: string,
+  indicatorColor: string,
+  bufferColor: string,
+  timeArr: number[],
```
Vanals, line 19: If there are utils only used by seek-bar.tsx can we move them in a `seek-bar/utils.ts` ?

src/audio-player-composable/audio-player-composable.tsx
Xin00163: As discussed, I have removed trackPositionArr and use trackPosition instead. For the values in slider, will have to do `[trackPosition]`

Vanals: Hope this change won't break anything 🤔 . Not sure if what you just mentioned in was the only reason why had it like that. Hopefully, tests should highlight any issues!

Xin00163: Will double check but the reason is because slider's value property is taking an array as you might have two thumb positions.

Vanals: I see! Not in our audio player tho, right? we won't have 2 thumbs

Xin00163: No I don't think so

src/audio-player-composable/context.tsx
```diff
@@ -35,8 +35,8 @@ interface AudioPlayerProviderContext {
   };
   getSeekBarProps: () => {
     duration: number;
-    trackPositionArr: number[];
-    onChangeSlider: ([value]: number[]) => void;
+    trackPosition: number;
+    onChange: ([value]: number[]) => void;
```
mutebg, line 39: can we change onChange to take a single value as well?
and on the Slider component, you can do something like
```
onChange={ ([value]) => onChange(value) }
```

Xin00163: Absolutely. Good spot!

mutebg (approved):

Vanals (approved):

PR #119, opened 2022-03-17
feat(PPDSC-2009): add reduced motion to transitionPreset

Xin00163:
PPDSC-2009

- Refactor `getTransitionPresetFromTheme` so that it will render both `no-preference` and `reduce` media queries.
- Change transitionDuration to 0 if motion is reduced

**I have done:**

**I have tested manually:**

**Before**

**After:**

**Who should review this PR**

**How to test**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2009-reduced-motion 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2009-reduced-motion/storybook

mutebg (changes requested):

src/utils/__tests__/get-transition-duration.test.ts
```diff
@@ -1052,4 +1052,298 @@ describe('getTransitionDuration', () => {
...
+      undefined,
+    )({
+      theme,
+    });
+
+    expect(result).toEqual({
+      '@media screen and (max-width: 799px)': {
+        appear: 2000,
+        enter: 2000,
+        exit: 2000,
```
mutebg, line 1111: I don't think that's the right behaviour, the way we use getTransitionDuration is inside JS ( React components )  to calculate the Transitions Timeout, so the function should just return a simple JS object or a number.  When you have MQ object it should just return the max values
@pdimova am I correct?

Xin00163: I think you are right. The returned value is for the timeout which won't recognise MQ

src/utils/get-transition-duration.ts
```diff
@@ -109,14 +111,43 @@ export const getTransitionDurationFromTheme = (
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      (acc, [bp, transitionPresetToken]) => {
+        let currentDuration;
+        if (isArrayLikeObject(transitionPresetToken)) {
+          const tokensArray = Object.values(transitionPresetToken) as string[];
+          currentDuration = extractDurationFromPreset(tokensArray, props.theme);
+        } else {
+          const tokensArray = [transitionPresetToken];
+          currentDuration = extractDurationFromPreset(tokensArray, props.theme);
+        }
```
mutebg, line 140: This is a bit of personal preference, you can write it like so:
```suggestion
      (acc, [, transitionPresetToken]) => {
        const tokensArray: string[] = isArrayLikeObject(transitionPresetToken)
          ? transitionPresetToken
          : [transitionPresetToken];

        const currentDuration = extractDurationFromPreset(
          tokensArray,
          props.theme,
        );
```

mutebg (dismissed):

src/utils/get-transition-duration.ts
Xin00163: This looks neat 🔥. Thanks!

mutebg (dismissed):

mutebg (approved):

baburay23 (approved):

PR #126, opened 2022-03-21
feat(PPDSC-2007): New Release flow and Gitflow to GitHub flow

Vanals:
PPDSC-2007

**What**

I have refactored the CircleCi Workflow. We moved from one single workflow to 4 different workflows. 
This will make our pipeline tidier and easier to read or to modify in the future.

- One workflow, `pr`, should trigger on any branch named `feat/, bugfix/ or docs/`.
- `main` on any main merge/commit.
- `on_release_tag` Triggered by release tag, we will be using the GitHub interface for doing so. It takes care of bumping the package.json version and updating main.
- `on_deploy_tag` will deploy main to prod after the bump has happened.

The refactoring has been done accordingly to the new GitHub Flow approach. So we won't be using any dev branch anymore or staging environment.  
I will be updating the release documentation as I do a release once merged.

**I have done:**

**I have tested manually:**

**Before**

**After:**

**Who should review this PR**

**How to test**

Release flow, a big picture.
release-process


.circleci/config.yml
```diff
@@ -872,139 +848,91 @@ jobs:
...
-
-# ========== REQUIREMENTS END   ==========
-
 workflows:
-  build_and_deploy:
+
+  pr:
+    when:
+      matches:
+          pattern: '^feat\/.*|bugfix\/.*|docs\/.*'
```
baburay23, line 874: just wondering if on the branch there is no` feat/bugfix `or` docs `name...does the workflow not run at all?Does the developer have to correct the branch name in order for it to run?

Vanals: That's correct! Before I was just checking if it wasn't `main` but this pattern feels more explicit and explains exactly when it runs.
Do we think might be a problem?
If I am not wrong those prefixes might also be used by a following enhancement for improving the release process. Which will check the branches merged into main.

baburay23: thanks Marco makes sense to be more explicit as opposed to just saying not main

.circleci/config.yml
```diff
@@ -1196,126 +1127,103 @@ workflows:
...
-          tf_env: false
-          requires:
-            - init_aws
-      - deploy_terraform:
-          <<: *only_develop
-          name: deploy_terraform_s3_dev
-          package: s3
+      # end
+
+  on_release_tag:
```
baburay23, line 1143: sorry i just wanted to make sure about the steps.
So for now we will manually add a release tag via the github UI?
Then it will run the bump version commands in the makefile which updates the version and tags it as a release and updates main. So does this mean when we add a release tag, the rest is done via the workflow and thats it?

Vanals: Correct! That's the idea!. We will be merging automatically the package json and then tell to circleci, via deploy tag, to start the deployment workflow.

.circleci/config.yml
jps: I think we should support all of the above aguably it's going to be easier to just do !main then you don't need to worry about it?

Vanals: deal 👍  -  I will add those

Vanals: Implemented. 

I was thinking, @jps , will this allow people to merge a PR without running any circleCI workflow if they don't fit in the naming illustrated in the regex? Will GitHub allow that?
If yes, maybe we want to make circleCi fail if they are working on a branch that is NOT either Main or fit the PR regex?.
Probably a workflow ad hoc would be needed for that.

Vanals: Just tested it with a custom branch name. Apparently won't be possible as you can't push if you branch name does not fit one of those types:

```[build, chore, ci, docs, feat, fix, improvement, perf, refactor, revert, style, test]```

Didn't know we had this. I can update this script to match the branch types we just agreed on above.


Screenshot 2022-03-23 at 16 59 44

File: .circleci/config.yml [hunk 3]
```diff
@@ -872,139 +848,92 @@ jobs:
...
-# ========== REQUIREMENTS END   ==========
-
 workflows:
-  build_and_deploy:
+
+  pr:
+    when:
+      matches:
+          # Supporting Pr branches starting with feat/, docs/, fix/, chore/, style/, refactor/, perf/, test/
+          pattern: '^feat\/.*|docs\/.*|fix\/.*|chore\/.*|style\/.*|refactor\/.*|perf\/.*|test\/.*'
```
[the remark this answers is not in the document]
Vanals: Yes we could! Initially I had  something like
```
when:
  not:
    - equal: [ main, branchName]
```

I just thought the above condition is not telling much. It is saying when NOT to run but not WHEN exactly is going to run. 
Just thought would be more explicit and clear stating when is going to run.
But if we feel is too wordy we can just negate `main`.

.circleci/config.yml
Vanals: @jps 

Uhm! I see what you mean.

Then we can  change slightly the way we do things so..

1) Trigger the release through GitHub UI with a `trigger-release@5.0.0` instead. So a trigger release tag rather than a release tag.

2) The on_trigger_release workflow will kick in, bump the package.json version and merge to main.
    THEN the same workflow creates the release tag as we know, `e.g: v.5.0.5` and pushes it

3) This will trigger the release deployment to Staging and PROD... PLUS will delete the trigger-release@5.0.0 tag.


In this way we will end up with a more clean tag history, just one (v.5.0.0) AND it should refer to the version bump in the package json.

.circleci/config.yml
```diff
@@ -24,7 +20,7 @@ cypress_defaults: &cypress_defaults
 applitools_defaults: &applitools_defaults
   command: |
     export APPLITOOLS_BATCH_ID=`echo $CIRCLE_SHA1`
-    export APPLITOOLS_PARENT_BRANCH_NAME=develop
+    export APPLITOOLS_PARENT_BRANCH_NAME=main
```
Xin00163, line 23: This is going to be interesting because we currently run snapshots against develop baseline. The first few runs might needs to be approved carefully

Vanals: Yes the idea is to change the base once this gets merged 👍

.circleci/config.yml
jps: Maybe either @SZYanev or @tbradbury might have be able to provide some input on the best way to do this? as I'd imagine that the render project has to do something similar.

Vanals: I have only managed to have a chat with Tim at the beginning of the ticket! Render team does handle things similarly but not quite. One of the main differences is that they do deploy EVERY TIME they merge to the main.
We do not want that (so far), we want to be able to choose when to release, that's why we kind of ended up with two tags; one for triggering the bump and one for deploying, they just create one after the package.json bump!

I have anyway double-checked with Tim this morning, explaining to him the concern you've raised and the solution I've proposed; he thinks it makes sense and doesn't see any issue with it!

If we are happy with this solution I can push it!
I think is good!

Vanals: Thank you @SZYanev!

At this point, @jps,  depends on how we want the release to start. Initially, we said we wanted to have it triggered by the GitHub UI interface? and also fill in the notes from there.

So if we are still happy with that Idea, I think using GitHub UI does still make sense. Is just another way of triggering the pipeline IMO.
Do you see any issue with that @SZYanev ? Also if we would do it from CircleCI I guess we will also need to implement the `release-documentation-cli`, which we wanted to explore on a following ticket.

SZYanev: No issue for me if you decide to go through the UI. And on the second note, yes you will need release-cli integrated if you automate the proccess.

.circleci/config.yml
Vanals: Updated regex with the pr branches types git supports by default 👍

newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2007-git-flow-to-github-flow 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2007-git-flow-to-github-flow/storybook

.circleci/config.yml
Vanals: release-process

Vanals: release-process

jps (approved):

Xin00163 (approved):

Xin00163 (approved):

Xin00163 (approved):

jps (approved):

PR #140, opened 2022-03-30
feat(PPDSC-2052): keyboard shortcuts next prev

mutebg:
PPDSC-2052

**What**

1. Background - add keyboard shortcut to skip-prev, skip-next, forward and replay buttob
2. What did you do -
     - update `useKeypress` hook to support multiple key presses at the same time, and not allowing the same keyboard shortcut to be attached more than 1 time.
     - create `useKeyboardShortcutsOnButton` hook and moves all keyboard logic to one place.
     - update all buttons to use common overrides hook `useButtonOverrides` so that logic is shared
     - add and update types where is needed 

**I have done:**
 - [x] Written functional tests against the component and/or NewsKit site

**I have tested manually:**
 - [x] The feature's functionality is working as expected on other browsers such as Safari, Firefox and IE11

**Before**

**After:**

**Who should review this PR**

**How to test**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2052-keyboard-shortcuts-next-prev-etc 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2052-keyboard-shortcuts-next-prev-etc/storybook

src/audio-player-composable/utils.ts
```diff
@@ -25,3 +33,39 @@ export const getMediaSegment = (duration: number, currentPosition: number) => {
...
+};
+
+export const useKeyboardShortcutsOnButton = ({
+  props,
+  defaults: defaultKeys,
+  action,
+}: UseKeyboardShortcutsType) => {
+  const {audioSectionRef} = useAudioPlayerContext();
+
+  const {keyboardShortcuts: keyboardShortcutsProp} = props;
```
Vanals, line 50: could this be kept as `keyboardShortcuts` ?  `keyboardShortcutsProp` sounds like a Typescript interface or type and was a bit confusing

src/audio-player-composable/__tests__/audio-player-composable-e2e.stories.tsx
```diff
@@ -23,62 +25,77 @@ const fullAudioPlayerAreas = `
...
+                      showEvent('skip-next');
+                    }}
+                  />
+                </GridLayout>
+              </Areas.Controls>
+            </>
+          )}
+        </GridLayout>
+      </AudioPlayerComposable>
+      <div data-testid="event">{event}</div>
```
Xin00163, line 96: Good idea!

src/audio-player-composable/types.ts
```diff
@@ -23,6 +28,47 @@ export interface AudioFunctionDependencies {
   setBuffered: React.Dispatch<React.SetStateAction<TimeRanges | undefined>>;
 }
 
+export interface AudioPlayerProviderContext {
```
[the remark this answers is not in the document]
mutebg: no, because we have circle dependency and that was the only way I can fix it.

src/audio-player-composable/audio-functions.ts
```diff
@@ -205,18 +205,6 @@ export const useAudioFunctions = ({
...
-  const onPopoutClick = () => {
-    pause();
-
-    fireEvent(
-      // TODO remove ignore when implementing live functionality and test
-      /* istanbul ignore next */
-      getTrackingInformation('audio-player-popout', EventTrigger.Click),
-    );
-  };
-
```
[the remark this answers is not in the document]
mutebg: We will not have plans for that, everyone can add this by themself, I will update storybook to open the link in a popup but in general is quite easy to implement

src/audio-player-composable/components/types.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import {IconButtonProps} from '../../button/types';
+import {Optional} from '../../utils/types';
+
+export type AudioPlayerIconButtonProps = Optional<
```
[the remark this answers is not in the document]
mutebg: I don't think so, every component has its own type like `AudioPlayerForwardButtonProps` and we export that, this is more like private one that is needed to build the rest

src/audio-player-composable/types.ts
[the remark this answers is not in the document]
mutebg: good one, I think we need, because the plan was to export `useAudioPlayerContext` which will allow people to create their own components if that's needed ( I saw that we forgot to export it so I will update it )

Xin00163 (dismissed):

src/audio-player-composable/components/time-display/utils.tsx
```diff
@@ -13,7 +13,10 @@ export const calculateTime = (secs: number) => {
   return `${returnedMinutes}:${returnedSeconds}`;
 };
 // change to formattedcurrentime
```
Vanals, line 15: is that a left over from someone 🤔 ?

src/utils/hooks/use-keypress.ts
```diff
@@ -10,11 +11,86 @@ interface Options {
 const isKeyboardEvent = (event: Event): event is KeyboardEvent =>
   'key' in event;
 
+/*
+ This store together with addEventIfNotExist and removeEvent track if the event wit the same
```
Vanals, line 15: lil typo, `wit` to `with`

src/audio-player-composable/audio-functions.ts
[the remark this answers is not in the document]
mutebg: I spoke with Mike about it, what I did is added this popup to the link, I guess everyone can add the tracking the same way we do for the rest of the buttons.
One thing regarding the tracking that I noticed is, right now skip next and previous does not have any tracking and that's the same for the old player. Do we need this tracking for all buttons or for none of them?

src/utils/hooks/use-keypress.ts
```diff
@@ -25,36 +101,64 @@ export const useKeypress = (
     const {enabled, eventType, target, preventDefault} = options;
 
     const handle: EventListener = (e: Event) => {
+      /* istanbul ignore if */
       if (!isKeyboardEvent(e)) {
         return;
       }
       if (e.defaultPrevented) {
         return;
       }
```
Vanals, line 110: which scenario is this covering 🤔 ?

src/utils/hooks/use-keypress.ts
```diff
@@ -10,11 +11,86 @@ interface Options {
...
+
+When the keyStore values match the key argument , the callback is called.
+https://www.gavsblog.com/blog/detect-single-and-multiple-keypress-events-javascript
+*/
 export const useKeypress = (
   key: string | string[],
   action: ((e: KeyboardEvent) => void) | undefined,
   opts?: Options,
 ) => {
+  const keyStore = useRef<{[key: string]: boolean}>({});
```
Vanals, line 88: Why are we using a ref, as a store? and not an Object? for example

src/utils/hooks/use-keypress.ts
```diff
@@ -25,36 +101,64 @@ export const useKeypress = (
...
+    const addToStore = (e: KeyboardEvent) => {
+      keyStore.current[e.key] = true;
+    };
+
+    const removeFromStore = (e: KeyboardEvent) => {
+      delete keyStore.current[e.key];
+    };
+
+    const onKeyDown = composeEventHandlers([addToStore, handle]);
+    const onKeyUp = composeEventHandlers([handle, removeFromStore]);
```
Vanals, line 142: Just to understand, why are we adding and removing it to the store on key down and key up 🤔 ?

src/utils/hooks/use-keypress.ts
mutebg: the `ref` is created only one time, and it's been reused over hook re-renders.
if I do `const keyStore = {}` that will create a new keyStore every time the hooks is executed even tho is the same instance. I can do useEffect and create the object there but I believe using a ref is the better approach.

Vanals: thanks

src/utils/hooks/use-keypress.ts
mutebg: imagine you want to catch up "ctrl + p" shortcut.
the keyboard events that will be triggered are the following:
- keydown with shift
- keydown with p
- keyup with shift
- keyup with p

what we want to catch is when shift and P are pressed at the same time, that's why on keydown the key is been added to the list and stored and on the keyup removed.
when the keys inside the store match "shift + p" we execute the action.

src/audio-player-composable/components/play-pause-button/play-pause-button.tsx
```diff
@@ -58,13 +38,17 @@ const ThemelessAudioPlayerPlayPauseButton: React.FC<PlayPauseButtonProps> = Reac
       [togglePlay],
     );
 
-    useKeypress(keyboardShortcuts.toggle, toggleAction, options);
+    useKeyboardShortcutsOnButton({
+      props: propsFromContext as AudioPlayerPlayPauseButtonProps,
+      defaults: defaultKeyboardShortcuts,
+      action: toggleAction,
```
Vanals, line 44: Could we add a lil comment about what the action is? I understood it only seeing the play pause button! I think will help

src/audio-player-composable/audio-functions.ts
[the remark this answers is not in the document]
Vanals: mmm you mean they will need to add getTrackingInformation them self? are we exporting it, right 🤔 ?
Hopefully we have a guide too.

MMm good question. This was implemented a years ago. Not sure if there was any reason for not adding it in the skip buttons? maybe there was no requirements. But I don't see why should not be there. James might remember

Vanals (approved):

Vanals (approved):

Xin00163 (approved):

Vanals (approved):

Xin00163 (approved):

PR #160, opened 2022-04-13
feat(PPDSC-2054): add feedback element and transition preset to slider, checkbox & radio button

evgenitsn:
PPDSC-2054

**What**

Added:
- Feedback element to the **thumb** on the **slider component**
- `opacityChange` preset
- `opacityChange` applied on the feedback element on the slider thumb
- `opacityChange` applied on the checkbox and radio button (base-switch) feedback elements
- Support for **active** state (mouse down event) for checkbox, radio button and slider thumb

**Feedback preset is now extracted and reused in the components mentioned above.**

This PR combines:
https://nidigitalsolutions.jira.com/browse/PPDSC-2054
and
https://nidigitalsolutions.jira.com/browse/PPDSC-1965

**I have done:**

**I have tested manually:**

https://user-images.githubusercontent.com/7677924/163189005-1c9ee95e-6520-44f4-ae31-7195cecf4059.mov

https://user-images.githubusercontent.com/7677924/163189012-d1725410-c1fb-4144-ad4a-1ccabadf6473.mov

https://user-images.githubusercontent.com/7677924/163189014-d5d9e8f8-79df-4146-adfb-05e0b7f4c6ab.mov


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2054-add-feedback-element-and-transition-preset-to-slider-checkbox-radio 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2054-add-feedback-element-and-transition-preset-to-slider-checkbox-radio/storybook

src/base-switch/base-switch.tsx
```diff
@@ -61,13 +66,30 @@ export const BaseSwitch = React.forwardRef<HTMLInputElement, BaseSwitchProps>(
...
+        }
+        setIsInputFocused(true);
+      },
+      [setIsInputFocused],
+    );
 
     const onInputBlur = useCallback(() => {
+      if (isInputFocusedVisible) {
+        setIsInputFocusedVisible(false);
+      }
```
mutebg, line 82: I wonder if we need this if statement here? isn't it the same?
```suggestion

        setIsInputFocusedVisible(false);
      
```

src/base-switch/base-switch.tsx
```diff
@@ -61,13 +66,30 @@ export const BaseSwitch = React.forwardRef<HTMLInputElement, BaseSwitchProps>(
...
     );
 
-    const onInputFocus = useCallback(() => {
-      setIsInputFocused(true);
-    }, [setIsInputFocused]);
+    const onInputFocus = useCallback(
+      (e: React.FocusEvent<HTMLInputElement>) => {
+        if (isFocusVisible(e)) {
+          setIsInputFocusedVisible(true);
+        }
```
mutebg, line 73: is it the same if we do 
```suggestion
   setIsInputFocusedVisible(isFocusVisible(e)); 
```

mutebg (approved):

src/base-switch/base-switch.tsx
evgenitsn: Yup, nice one.

src/base-switch/base-switch.tsx
evgenitsn: It should be the same yes. Just tested that.

mutebg (approved):

src/base-switch/base-switch.tsx
```diff
@@ -115,6 +134,8 @@ export const BaseSwitch = React.forwardRef<HTMLInputElement, BaseSwitchProps>(
         {...(label ? labelAttributes : {})}
         onMouseOver={onLabelMouseOver}
         onMouseLeave={onLabelMouseLeave}
+        onMouseDown={onMouseDown}
```
pdimova, line 137: can we give this container a more descriptive name, my suggestion would be `StyledSwitchAndLabelWrapper` or something similar

src/slider/styled.ts
```diff
@@ -150,6 +152,29 @@ export const StyledThumb = styled.div<StyledThumbProps>`
   cursor: ${getCursor};
 `;
 
+export const StyledFeedback = styled.span<
```
pdimova, line 155: What is the reason for having another StyledFeedback instead of reusing the one in BaseSwitch? 

Also, can we give the two Feedback components more descriptive names
StyledSwitchFeedback and StyledThumbFeedback

src/base-switch/base-switch.tsx
```diff
@@ -142,6 +163,7 @@ export const BaseSwitch = React.forwardRef<HTMLInputElement, BaseSwitchProps>(
             overrides={overrides}
             size={size}
```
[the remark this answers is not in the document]
evgenitsn: Correct, we should not trigger the feedback on focus. The focus state would trigger a browser default focus ring around the element and that can change in the near future because there is a spike to assess focus state through all of the elements.

src/base-switch/styled.tsx
```diff
@@ -85,6 +87,13 @@ export const StyledSwitch = styled.div<
       isFocused,
       isHovered,
     })};
+
+  ${({isFocusedVisible}) =>
```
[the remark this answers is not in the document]
evgenitsn: Yes we would use the default browser support for focus for now.

src/slider/styled.ts
evgenitsn: For the first part TBH maybe I am not the best person to answer because I don't have the background of implementing these but what I can see is that both Feedback elements have some differences in terms of specifying the style presets path and also their underlying interactive element is different. 
`BaseSwitch` is depending on an `<input/>` and `Slider` is depending on the 3rd party `react-range` library which uses a `<div/>` for an interactive element.

Not sure if we can generalise and extract that to a single component. 
Also not sure where should live if it's a separate piece.

src/slider/slider.tsx
```diff
@@ -117,6 +118,11 @@ const ThemelessSlider: React.FC<SliderProps> = ({
         data-testid={`${dataTestId}-thumb`}
         overrides={overrides}
       >
+        <StyledFeedback
```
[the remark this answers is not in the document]
evgenitsn: Actually I am not sure about that one. I haven't thought about it at all. 
Now looking at the code I can see that we control with layering on the BaseSwitch using this right?
```
const STACKING_CONTEXT = {
  feedback: '1',
  input: '2',
};
```
I haven't seen any specific requirement for that, so maybe we want to align that layering for the slider to be the same as the BaseSwitch?

src/utils/focus-visible.ts
```diff
@@ -0,0 +1,11 @@
+export const isFocusVisible = (e: React.FocusEvent<HTMLInputElement>) => {
+  try {
+    return e.target.matches(':focus-visible');
+  } catch (error) {
+    /*
+      Currently only Safari 15.4 supports :focus-visible.
+      For the previous versions of Safari, we will return true here which means that the focus ring on the elements which are using this function will be visible on mouse click. So this acts more like an incremental enhancement for Safari. For all other supported browsers this will work as expected because they have good support for :focus-visible.
+    */
+    return true;
```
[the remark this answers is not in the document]
evgenitsn: False would be returned only in the `try` block if the element doesn't have this selector applied, i.e. if this is false `e.target.matches(':focus-visible');`

In the `catch` block the result will only be true. The `catch` block will only be triggered if the browser does not support `:focus-visible` which means that in this case it's better to just `return true` to toggle the focus ring for the targeted element every time.

src/slider/slider.tsx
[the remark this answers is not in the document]
pdimova: My personal preference would be to get the Feedback component to work the same way across all components, including the layering as well. 
But on the other hand, it looks like there is no need for such change now. 

@mutebg We need to check how these z-indices will work inside a Layer for example

pdimova (approved):

tbradbury (approved):

PR #167, opened 2022-04-19
chore(PPDSC-2079): update axe-core and fix any errors

evgenitsn:
PPDSC-2079

**What**

Our `axe-core` and `cypress-axe` versions were not up to date, so I've updated them and cleared all new a11y errors.

**I have done:**

**I have tested manually:**

**How to test:**
Run `yarn e2e:build` then `yarn e2e:comps` and check if all tests pass (but that should be handled by circleci either way).


src/scroll/scroll.tsx
```diff
@@ -122,7 +122,6 @@ const ThemelessScroll = React.forwardRef<HTMLDivElement, ScrollProps>(
           tabIndex={0}
           {...props}
           vertical={vertical}
-          aria-orientation={vertical ? 'vertical' : 'horizontal'}
```
evgenitsn, line 125: This is not relevant here and the axe-core is throwing errors for not valid attribute. 
The `aria-orientation` should be one of these components (widgets) https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-orientation#description

newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2079-update-axe-core-and-fix-any-errors 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2079-update-axe-core-and-fix-any-errors/storybook

Srinivasan-Ramasamy (dismissed):
The code changes look good to me :+1: let's wait others point

mutebg (commented):
👏

mutebg (approved):

site/components/__tests__/__snapshots__/site-header.test.tsx.snap
```diff
@@ -1151,7 +1151,6 @@ exports[`SiteHeader has first item highlighted (About) if we are on a section wi
               class="emotion-56"
             >
               <div
-                aria-orientation="horizontal"
```
Vanals, line 1154: Why was this attribute removed? 
I wonder if this should be kept and added a role='menu' instead

Vanals (changes requested):

src/scroll/scroll.tsx
Vanals: Rather than remove it, should we instead add the `scrollbar` role? 
Seems like that's what the component is

src/slider/slider.tsx
```diff
@@ -78,9 +78,6 @@ const ThemelessSlider: React.FC<SliderProps> = ({
         dragged={isDragged}
         disabled={disabled}
         vertical={vertical}
-        aria-valuemin={min}
-        aria-valuemax={Math.floor(max)}
-        aria-orientation={vertical ? 'vertical' : 'horizontal'}
```
Vanals, line 83: have they been removed for the same reason mentioned above? I wonder if the same can be done.

Vanals (changes requested):

src/scroll/scroll.tsx
evgenitsn: Adding a `role='scrollbar'` then throws another error for not having `aria-valuenow` attribute. 
image

In this article are described all required attributes for a scrollbar https://www.digitala11y.com/scrollbar-role/
I am not sure if the intent of a `scrollbar` is the same as our use case for the `scroll` component. 
Also the children of the `scroll` component are dynamic (up to the consumer) and they can be a single child with many children inside (like a Scroll with Stack inside) or multiple children (like a set of images). These examples are from storybook. So we can't be sure if the direct children of the Scroll would be a single element or many elements.

src/slider/slider.tsx
evgenitsn: So these have been moved from the track to the thumb element of the same component. 
axe-dev is showing this error and after a research I found out that these aria attributes should be on the thumb so I moved them and the errors are now gone.

image

src/scroll/scroll.tsx
Vanals: uhm! Ok! thanks! Let see if we get any feedback in future!

src/slider/slider.tsx
Vanals: Make sense! thanks! I am just bit confused because if I go in storybook prod or dev I can see these values are already in the thumb?
Also, does the screen reader reads ok after this change? Not sure what scenario  was affected if any specificly

Vanals (approved):

Srinivasan-Ramasamy (approved):

mutebg (approved):

PR #182, opened 2022-04-28
feat(PPDSC-2078): switch v1

mstuartf:
[PPDSC-2078](https://nidigitalsolutions.jira.com/browse/PPDSC-2078)

**What**

1. Make required changes to BaseSwitch to support new Switch component
2. Make sure Checkbox and RadioButton interfaces are unchanged
3. Add v1 Switch component meeting a11y requirements with snapshot tests

**Before:**
n/a

**After:**
Screenshot 2022-04-28 at 12 36 45
Screenshot 2022-04-28 at 12 36 59
Screenshot 2022-04-28 at 12 37 10
Screenshot 2022-04-28 at 12 37 20

**How to test:**
Review in Storybook


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2078-switch-v1 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2078-switch-v1/storybook

src/switch/switch.tsx
```diff
@@ -0,0 +1,75 @@
...
+  checked,
+  parentOverrides,
+  children,
+  isFocused,
+  isHovered,
+}: BaseSwitchIconProps & {children?: React.ReactNode}) => {
+  const overrides = parentOverrides as SwitchOverrides;
+  const {onIcon: OnIcon, offIcon: OffIcon} = overrides;
+  return (
+    <StyledSwitchContainer size={size} overrides={overrides} aria-hidden>
```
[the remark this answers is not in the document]
mstuartf: My thinking here was that users can pass in arbitrary components as icons, which could contain text. So I added aria-hidden here to stop them being read in that case. 

But I think perhaps that's not a case we really need to guard against. I'll remove.

src/base-switch/styled.tsx
```diff
@@ -104,14 +120,37 @@ export const StyledSwitch = styled.div<
...
 >`
   position: absolute;
   top: 50%;
-  left: 50%;
+
+  ${({centreOnThumb, size, checked, path, ...rest}) => {
+    if (!centreOnThumb) {
+      return {left: '50%'};
+    }
+    const {padding} = getResponsiveSize(
```
mutebg, line 136: small problem with that, this will not always return a `{padding: 'value'}` object, since customers can set media query object for spaceInset, in that case it will return 
```
{
'@media screen and (max-width: 479px)': {padding: 'Xpx'},
 '@media screen and (min-width: 480px) and (max-width: 767px)': {padding: 'Ypx'},
 '@media screen and (min-width: 768px)': {padding: 'Zpx'},
}
```

src/base-switch/types.ts
```diff
@@ -18,16 +23,27 @@ type BaseSwitchOverrides = {
     stylePreset?: MQ<string>;
     spaceInline?: MQ<string>; // Note: spaceInline is used as a gap. Should rename this instead of using logical props.
     transitionPreset?: TransitionToken | TransitionToken[];
+    blockSize?: MQ<string>;
+    inlineSize?: MQ<string>;
+    spaceInset?: MQ<string>;
```
mutebg, line 28: Maybe stop using spaceInset and instead add paddingBlock and Inline and here so that we don't have to update it later.

src/base-switch/types.ts
```diff
@@ -18,16 +23,27 @@ type BaseSwitchOverrides = {
...
     typographyPreset?: MQ<string>;
   };
+  thumb?: {
+    size?: MQ<string>;
+    stylePreset?: MQ<string>;
+    transitionPreset?: TransitionToken | TransitionToken[];
+  };
   icon?: Override<BaseSwitchIconProps>;
+  onIcon?: Override<BaseSwitchIconProps>;
+  offIcon?: Override<BaseSwitchIconProps>;
```
mutebg, line 46: Do you think BaseSwitch needs these 2 or they can be available only for Switch

src/theme/presets/transition-presets.ts
```diff
@@ -266,3 +266,11 @@ transitionPresets.growOnClick = {
     transitionDuration: '{{motions.motionDuration020}}',
   },
 };
+
+transitionPresets.shiftAbsolute = {
+  base: {
+    transitionProperty: 'inset',
```
mutebg, line 272: Good one 👏

mutebg (commented):
Congrats on your first component, well done, few small comments

src/base-switch/styled.tsx
mstuartf: Good catch. Is there a way to get the applied padding value or does the approach need a rethink?

src/base-switch/types.ts
mstuartf: Yes, good point!

src/base-switch/styled.tsx
mutebg: I don't think there is a way around it. Probably need to rethink that one :/

mutebg: My suggestion is to set `box-sizing: border-box;` on the element, in that case,  the padding will not change the size of the element, and it will be inside, it is quite often when we use padding also to use border-box so that elements keeps it sizes.

mstuartf: As discussed, I've updated this to directly measure the padding of the container and pass this to the feedback element to use as its offset.

src/base-switch/types.ts
mstuartf: It does create typing issues if they are only on the Switch interface.

src/switch/switch.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {SwitchIconProps, SwitchProps} from './types';
+import defaults from './defaults';
+import stylePresets from './style-presets';
+import {withOwnTheme} from '../utils/with-own-theme';
+import {BaseSwitch} from '../base-switch';
+import {StyledSwitchContainer, StyledThumb, StyledTrackColumn} from './styled';
+import {BaseSwitchIconProps} from '../base-switch/types';
+import {Override} from '../utils/overrides';
+
+// TODO: Contains similar logic to getComponentOverrides to cast the override icon as a component. But why is this necessary?
```
mutebg, line 12: Here is a bit controversial,
we don't have default off-Icon and on-Icon so in that sense why are they in overrides? 

the idea of using getComponentOverrides is to be able to override a specific component in  3 ways:
1: What we have for most of the components, you can do that just as passing stylePreset and and other style-related things:
```
overrides: {
    scroll: {
        stylePredet: 'newStylePreset',
    }
}
```
2: Override or add a new prop to that component, there are some cases that user want to change not only styles but also some of the props of underneath components.
```
overrides: {
    scroll: {
        props:  {
           scrollBar: true
        }
    }
}
```
3: The user wants to replace the whole component with a new one
```
overrides: {
   scroll: ({children}) => <MyCustomScroll>{children}</MyCustomScroll>,
}
```
these examples are from the Tabs component.

In the case of `on and off icons` seems that we have only option 3 because we don't have a default on and off icons so your code seems fine to me

src/switch/styled.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }
+>`
+  display: flex;
+  width: 100%;
+  ${({justifyContent}) => ({justifyContent})};
+
+  ${({size, state}) =>
+    getStylePreset(`switch.${size}.column`, 'column', {
+      isDisabled: state === 'disabled',
+    })};
```
mutebg, line 32: I think it might be easy if we remove that part and add icon colors directly to `switchTrack` since it is the parent
```
switchTrack: {
    base: {
      backgroundColor: '{{colors.interactiveInput020}}',
      borderRadius: '{{borders.borderRadiusPill}}',
      iconColor: '{{colors.inkInverse}}',
    },
    checked: {
      backgroundColor: '{{colors.interactiveInput040}}',
    },
    disabled: {
      backgroundColor: '{{colors.interactiveDisabled010}}',
      iconColor: '{{colors.inkNonEssential}}',
    },
  },
```
in that case we also can delete `column` from the defaults

src/base-switch/types.ts
```diff
@@ -18,16 +23,27 @@ type BaseSwitchOverrides = {
     stylePreset?: MQ<string>;
     spaceInline?: MQ<string>; // LOGICAL_PROPS_TO_DO: Used as the gap between items. Should be renamed.
     transitionPreset?: TransitionToken | TransitionToken[];
+    blockSize?: MQ<string>;
+    inlineSize?: MQ<string>;
+    paddingInline?: LogicalPaddingProps['paddingInline'];
```
mutebg, line 28: What is the reason to have only paddingInline?

src/base-switch/styled.tsx
```diff
@@ -51,6 +53,47 @@ export const StyledSwitchContainer = styled.div<
...
+    )}
+
+  // If the blockSize is overridden to a low value, the Switch aligns above the
+  // label. We can't vertically align the Switch centrally because by
+  // design it is supposed to be aligned at the top (e.g. consider multi-line
+  // labels). Instead, use margins if the blockSize is smaller than the default. 
+  ${({size, path, ...rest}) => {
+    const {blockSize: defaultBlockSize} = getResponsiveSize(
+      'blockSize',
+      `${path}.${size}.input`,
```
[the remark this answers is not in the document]
mstuartf: So for this, I need to know: 

(a) if the blockSize has been overridden, and 
(b) if so, whether the new size is smaller than the default value for the Switch size in question

If both are true, I need to apply the margin. 

To do this I've called `getResponsiveSize` twice with the same path (`${path}.${size}.input`): 

* once _without_ the override path to get the default value (this should only be a single value not an MQ because it will not return an override, just the value for the size Switch in question from `defaults.ts`), and 
* once _with_ the override path (this will either be a single value or an MQ, but the callback should cope with both). 

Alternatively, is there a normal way to get the single default value without the first call to `getResponsiveSize`?

src/switch/switch.tsx
mstuartf: OK, thanks.

src/switch/styled.tsx
mstuartf: I originally had this but because the Thumb is a child of the Track, sometimes the Thumb icon colour would be overridden and set to the Track icon colour. The behaviour was very inconsistent and hard to debug. For example, if you added an override for `input.blockSize`, suddenly the Thumb icons would be the same colour as the Track icons. 

Adding a separate style preset for the Track icons in the `column` resolved the issue. 

What do you think?

src/base-switch/base-switch.tsx
```diff
@@ -33,6 +33,7 @@ export const BaseSwitch = React.forwardRef<HTMLInputElement, BaseSwitchProps>(
       path,
       defaultSwitchSelectorComponent: defaultSwitchComponent,
       type,
+      hideFeedback,
```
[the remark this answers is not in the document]
mstuartf: Sure, good point.

src/base-switch/types.ts
mstuartf: I wasn't sure of the use case but I guess there's no harm in adding support for all. Will update.

src/base-switch/base-switch.tsx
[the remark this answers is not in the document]
mstuartf: I'd actually not added support for this prop to the Checkbox by excluding the prop. I think this makes sense to avoid scope creep (ideally after this PR the behaviour of other components should be unchanged). 

Unless you disagree, I'd prefer to just exclude this prop from the RadioButton in the same way. If we want to add support for `hideFeedback` to those components, we can do so in a later ticket?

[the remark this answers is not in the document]
mutebg: We can do it on a later ticket. Can you create one please

src/base-switch/styled.tsx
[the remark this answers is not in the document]
mutebg: The problem I see is that people can override their defaults and so they can set MQ token as the default size for the Switch.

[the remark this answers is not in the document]
mstuartf: Do you mean if they set the overrides for the component at the theme level?

src/switch/styled.tsx
mutebg: I see, maybe we can find a better name for that? to me was not easy to understand what a `column` is doing

Xin00163 (commented):
There are some errors on the console in storybook:
`Error: <svg> attribute height: Expected length, "medium".`

src/switch/styled.tsx
Xin00163: can we call it track?

src/base-switch/styled.tsx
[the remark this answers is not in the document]
mstuartf: As discussed, overrides at the theme level change the defaults, so it's possible for defaults to be MQs. This approach is also problematic as there are too many customisation options that make it hard to reliably calculate the required offset. 

Instead we should allow the user to use marginBlock to manually adjust the Switch alignment if necessary. But the margin needs to be applied to the `StyledSwitchContainer` (not the `StyledSwitch`), so the logical props for `input` will need to be split. This seems fine as we already share the `input` styling between these two elements.

src/switch/types.ts
```diff
@@ -0,0 +1,53 @@
...
+    | 'transitionPreset'
+    | 'spaceInline'
+  > &
+    LogicalProps;
+  thumb?: BaseSwitchOverrides['thumb'];
+  feedback?: BaseSwitchOverrides['feedback'];
+  label?: BaseSwitchOverrides['label'];
+  onIcon?: BaseSwitchOverrides['onIcon'];
+  offIcon?: BaseSwitchOverrides['offIcon'];
+  icon?: BaseSwitchOverrides['icon'];
```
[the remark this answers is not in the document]
mstuartf: Good idea.

src/base-switch/base-switch.tsx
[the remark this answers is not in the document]
mstuartf: Ticket here: https://nidigitalsolutions.jira.com/browse/PPDSC-2168

Xin00163 (dismissed):

Xin00163 (commented):
Very nice work! Seems more complex than we originally thought but it's well-handled.

Xin00163 (approved):

tbradbury (approved):

PR #194, opened 2022-05-09
Fix/ppdsc 2113 prodtealium not being served 

Vanals:
PPDSC-2113

**What**

We were trying to access to `productionSiteEnv` through `this.props`. this.props.productionSiteEnv is undefined.

None of the props created at getStatiprops level were defined. The `production` prop, wasn't even used anywhere.
I have so cleaned the props, removing production and productionSiteEnv, then questioned the rest of the function. If renderPage is needed or if getStaticProps is needed at all.

I am then accessing the values, directly from `process` without passing them as a prop(which wasn't happening anyway). Time ago when we were using  `getInitialProps` the props were working, but we moved away from it for fixing an issue with HeadNext.

*Removing the whole function*

I could not find in the documentation any guidance around renderPage(), that could fit us, I could not find any guidance around why we should use getStaticProps in the _document. Instead, getStaticProps, is stated, is not supported in _document AND when building the page, after removing all of it, the website seems working fine.

https://nextjs.org/docs/advanced-features/custom-document
```
Document currently does not support Next.js [Data Fetching methods](https://nextjs.org/docs/basic-features/data-fetching/overview) like [getStaticProps](https://nextjs.org/docs/basic-features/data-fetching/get-static-props) or [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props).
```

[renderPage](https://nextjs.org/docs/advanced-features/custom-document#customizing-renderpage)

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


mutebg (approved):
thanks for explanation, seems good to me

baburay23 (approved):

PR #204, opened 2022-05-18
feat(PPDSC-2117): tooltip panel and content

Xin00163:
PPDSC-2117

- Build tooltip with https://github.com/floating-ui/floating-ui.
- Accessibility discussion: https://github.com/floating-ui/floating-ui/issues/1704
- Didn't add `onDismiss` or `onOpen` props because it seems unnecessary for tooltip. 
- Add React.forwardRef to icon button

Note 🚨: If the Tooltip is wrapping a functional component, the functional component needs to accept a ref using forwardRef. 

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2117-tooltip-panel-and-content 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2117-tooltip-panel-and-content/storybook

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+describe('Tooltip', () => {
+  const defaultProps = {
+    children: <button type="submit">Add</button>,
+    title: 'hello',
+    defaultOpen: true,
+  };
+
+  afterEach(() => {
+    cleanup();
```
baburay23, line 17: sorry can i ask why you are using cleanup() here? i looked it up and it says _Unmounts React trees that were mounted with [render]_  but i only see renderWithTheme() which is used elsewhere without a cleanup

Xin00163: This is the library's recommendation: https://floating-ui.com/docs/react-dom but I just commented it out and it passed as well 🤔

Xin00163: Having a closer look. I think it's because we are mocking the ResizeObserver so we don't need to clean up. Could get this removed. Without the mock, we will get an error `ResizeObserver is not defined`.

Xin00163 (commented):
Thinking about adding a warning message on prod env: 
```
Note 🚨: If the Tooltip is wrapping a functional component, the functional component needs to accept a ref using forwardRef.
```
Is it necessary?

src/icon-button/icon-button.tsx
```diff
@@ -6,7 +6,10 @@ import defaults from './defaults';
 import stylePresets from './style-presets';
 import {withOwnTheme} from '../utils/with-own-theme';
 
-const ThemelessIconButton = ({overrides = {}, ...props}: IconButtonProps) => {
+const ThemelessIconButton = React.forwardRef<
+  HTMLButtonElement | HTMLAnchorElement,
+  IconButtonProps
+>(({overrides = {}, ...props}, ref) => {
```
mutebg, line 12: Well done, someone had to do it :D

src/tooltip/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  placement?: Placement;
+  labelTooltip?: boolean;
+  overrides?: {
+    zIndex?: number;
+    maxWidth?: MQ<string>;
+    minWidth?: MQ<string>;
+    panel?: {
+      stylePreset?: MQ<string>;
+      typographyPreset?: MQ<string>;
+    } & LogicalPaddingProps;
```
mutebg, line 24: Why do we need the `panel` to me seems the same thing, setting min and max-width is also on the panel ( or technically on the parent ) but in the end, you are controlling the tooltip, if we don't plan to add more parts to that I think is better to have flat structure

src/tooltip/types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface TooltipProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title' | 'defaultValue'> {
+  children: React.ReactElement;
+  title: React.ReactNode;
+  open?: boolean;
+  defaultOpen?: boolean;
+  trigger?: TriggerType | TriggerType[];
+  placement?: Placement;
+  labelTooltip?: boolean;
```
mutebg, line 16: Can we rename that prop to something like `asLabel`.   `<Tooltip labelTooltip />` is the same as having `<Card cardTitle="" cardContent="" /> ` usually you don't want your props to include the name of the component and is better to be like `<Card title="" content "" />`

src/tooltip/styled.tsx
```diff
@@ -0,0 +1,23 @@
...
+  styled,
+  getResponsiveSpace,
+  getResponsiveSize,
+} from '../utils/style';
+import {logicalProps} from '../utils/logical-properties';
+
+export const StyledTooltipPanel = styled.div<Pick<TooltipProps, 'overrides'>>`
+  ${getStylePreset('tooltip.panel', 'panel')};
+  ${getTypographyPreset('tooltip.panel', 'panel')};
+  ${getResponsiveSpace('padding', 'tooltip.panel', 'panel', 'spaceInset')};
```
mutebg, line 14: Can we just use logical props here?

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+  ]);
+
+  // If tooltip is used as a label, add aria-labelledby to childrenProps and id to StyledTooltip;
+  // Because of above, 'aria-describedby' has different id for reference and floating, hence manually set below as well;
+  const id = useId();
+
+  const labelOrDescProps = {} as {
+    'aria-labelledby': string | null;
+    'aria-describedby': string | null;
+  };
```
mutebg, line 58: as far I can see you never have both properties at the same time
```suggestion
  const labelOrDescProps = {} as {
    'aria-labelledby'?: string | null;
    'aria-describedby'?: string | null;
  };
```

mutebg (commented):
Great job @Xin00163 👏  and all the examples in storybook are 💯

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  if (!title) {
+    return children;
+  }
+
+  return (
+    <>
+      {React.cloneElement(
+        children,
+        getReferenceProps({ref: reference, ...childrenProps}),
```
mutebg, line 79: Found a small issue with that reference, if the user has provided its own reference on the children it won't get any reference, since we are cloning the component with a new one
```
const ref = React.useRef(null);

  React.useEffect(() => {
    console.log(ref);
  }, [ref]);

  return (
    <>
      <Tooltip title="Add" placement="right" trigger={['focus', 'hover']}>
        <Button
          ref={ref}
          size={ButtonSize.Small}
          overrides={{stylePreset: 'buttonOutlinedPrimary'}}
        >
          btn
        </Button>
      </Tooltip>
    </>
  );
  ```
You will see that always console.log null.
  
What you can do is use `composeRefs` function ( we used it in many other places )
```
getReferenceProps({ref: composeRefs(reference, children.ref), ...childrenProps,}),
```
In that way, users will have their own reference to the children component

File: src/tooltip/types.ts (hunk 3 above)
Comment (Xin00163) [reply src/tooltip/types.ts:24]: That's what I thought as well. I had the discussion with Mike and James and decided to keep panel. I think maybe at that point we are not clear with what it will look like. Will have a chat with Mike around this again.

File: src/tooltip/styled.tsx (hunk 5 above)
Comment (Xin00163) [reply src/tooltip/styled.tsx:14]: Forgot to delete this line

File: src/tooltip/types.ts [hunk 8]
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import {Placement} from '@floating-ui/react-dom-interactions';
+import {MQ} from '../utils/style';
+import {LogicalPaddingProps} from '../utils/logical-properties';
+
+export type TriggerType = 'hover' | 'focus';
+
+export interface TooltipProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title' | 'defaultValue'> {
+  children: React.ReactElement;
```
mstuartf: It might be worth specifying here that the child component needs to accept a ref prop. Even if you specify it here I don't think the compiler would actually pick it up if the user passes in a component that doesn't have a ref prop. But I think having it detailed in the interface would be helpful to users anyway as the ref is important for the tooltip to work.

src/tooltip/tooltip.tsx
Xin00163: This is a good catch! I've updated and can see the ref coming through

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+  title,
+  placement = 'top',
+  trigger = 'hover',
+  open: openProp,
+  defaultOpen,
+  labelTooltip,
+  overrides,
+  ...props
+}) => {
+  const [open, onOpenChange] = useControlled({
```
mstuartf, line 31: Would `setOpenState` be a better name? It's more in line with other uses of `useControlled` (e.g. `BaseSwitch` and `RadioGroup`

src/tooltip/types.ts
Xin00163: Yes, good point, I've updated the types include ref. Also gonna update the docs to include a message like this `Note 🚨: If the Tooltip is wrapping a functional component, ensure that the functional component accepts a ref using forwardRef.`

src/tooltip/types.ts
Xin00163: Have removed panel here.

mstuartf (approved):

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,111 @@
...
+  title,
+  placement = 'top',
+  trigger = 'hover',
+  open: openProp,
+  defaultOpen,
+  asLabel,
+  overrides,
+  ...props
+}) => {
+  const [open, setOpenState] = useControlled({
```
Vanals, line 31: should this just be `setOpen` ? by convention

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,111 @@
...
+      enabled: trigger.includes('hover'),
+    }),
+    useFocus(context, {enabled: trigger.includes('focus')}),
+    useRole(context, {enabled: !asLabel, role: 'tooltip'}),
+    useDismiss(context),
+  ]);
+
+  // If tooltip is used as a label, add aria-labelledby to childrenProps and id to StyledTooltip;
+  // Because of above, 'aria-describedby' has different id for reference and floating, hence manually set below as well;
+  const id = useId();
```
Vanals, line 54: I think would be good mention to Mike the scenario of a button with an `emoji` as a text! So we can explain in the documentation when they should do as such 👍

src/tooltip/tooltip.tsx
Xin00163: I guess it's more concise

src/tooltip/tooltip.tsx
Xin00163: Yeah I will update the docs with more details

src/tooltip/tooltip.tsx
Vanals: I would stick to the convention of UseState, if applicable, is more readable: https://reactjs.org/docs/hooks-state.html

Xin00163:
> Looks good, just shared with you that issue with aria-label. Worth ticking the checkboxes once done, especially the one about Screen Readers! 😬 You can also test on NVDA with Assistive Labs

So as a result, we will need to make `aria-label` optional in Icon Button, otherwise we can't match the aria-label with the tooltip title.

src/tooltip/tooltip.tsx
```diff
@@ -53,13 +53,17 @@ const ThemelessTooltip: React.FC<TooltipProps> = ({
   // Because of above, 'aria-describedby' has different id for reference and floating, hence manually set below as well;
   const id = useId();
 
+  const titleIsString = typeof title === 'string';
```
Vanals, line 56: `isTitleString` I would personally go for `isTitleString`. That's the convention for boolean. is interrogative 👍

src/tooltip/tooltip.tsx
```diff
@@ -53,13 +53,17 @@ const ThemelessTooltip: React.FC<TooltipProps> = ({
...
+
   const labelOrDescProps = {} as {
+    'aria-label'?: string | null;
```
Vanals, line 59: As far as I know.I think this should be instead:

```
  const labelOrDescProps: {
    'aria-label'?: string | null;
    'aria-labelledby'?: string | null;
    'aria-describedby'?: string | null;
  } = {}
  ```
  
  As it is now, you are "forcing it to be as.. "

File: src/tooltip/tooltip.tsx [hunk 14]
```diff
@@ -0,0 +1,115 @@
...
+            id,
+            className: 'Tooltip',
+            style: {
+              position: strategy,
+              top: y ?? '',
+              left: x ?? '',
+            },
+          })}
+          overrides={overrides}
+          {...props}
```
Vanals, line 103: What about destructure and clearly pass the props here? I had to console log in order to see what was being passed to the tag

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,115 @@
...
+      {React.cloneElement(
+        children,
+        getReferenceProps({
+          ref: composeRefs(reference, children.ref),
+          ...childrenProps,
+        }),
+      )}
+      {open && (
+        <StyledTooltip
+          as={titleIsString ? 'p' : 'div'}
```
Vanals, line 91: The title could be a string or.. what? do we expect the use to pass also images inside? if not I would just keep it as a p

src/tooltip/tooltip.tsx
Xin00163: I don't mind. I did see `childrenIsString` and `testSvgCodeIsString` so don't think we are following this convention

src/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const labelOrDescProps = {} as {
+    'aria-label'?: string | null;
+    'aria-labelledby'?: string | null;
+    'aria-describedby'?: string | null;
+  };
+
+  if (asLabel) {
+    labelOrDescProps['aria-label'] = titleIsString ? title : null;
+    labelOrDescProps['aria-labelledby'] = open && !titleIsString ? id : null;
```
Vanals, line 66: Which scenarios are these covering? why do we need both? Maybe add some comments would be helpful

src/tooltip/tooltip.tsx
Xin00163: which prop?

src/tooltip/tooltip.tsx
Xin00163: The title is `React.ReactNode` in the spec so there is a chance it is not string.

src/tooltip/tooltip.tsx
Vanals: Is this gonna be done in a ticket? can we add it in the ticket if so?

Vanals:
Can we please use the checkboxes? and use N/a if not applicable? I think we should start getting use to those, if we want to keep them.. they would have also found the issue with the aria label.

src/tooltip/tooltip.tsx
Xin00163: There is a comment above but let me know if it's not clear

Vanals: I think we should, but ok 👍

src/tooltip/tooltip.tsx
Vanals: As discussed, is good 👍  no one of the one we declared are being drilled

src/tooltip/tooltip.tsx
Vanals: gotcha, thank you

src/tooltip/tooltip.tsx
Vanals: I see, unfortunately not. I think we should but ok 👍

Vanals (dismissed):

tbradbury (dismissed):

tbradbury (approved):

Vanals (approved):

PR #213, opened 2022-05-27
feat(PPDSC-1986): layout props 6/7

evgenitsn:
PPDSC-1986

**What**

- added margin and padding logical props support
- storybook scenario
- updated docs

for the following components:
`Tag`
`Tabs`
`Text Field`
`Title Bar`

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/site/pages/components/title-bar.mdx
+++ b/site/pages/components/title-bar.mdx
@@ -5,6 +5,7 @@ import {LegacyBlock} from '../../components/legacy-block';
 import {TextBoxSwatch} from '../../components/theming-values/colors';
 import {CodeFromDefaultPresets} from '../../components/code';
 import {Link} from '../../components/link';
+import {CommonLogicalPropsMDX} from '../../components/component-api/common-logical-props';
 
 import {TitleBar, Button, Headline} from 'newskit';
 
@@ -24,19 +25,24 @@ Title bar has a number of props to facilitate a variety of uses:
   <Prop name="children" type="string" required>
     The content of the title is passed as the child of the component.
   </Prop>
-  <Prop name="headingAs" type="'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span'" default="h3">
-    The heading tag used for the title. It should be one of the following strings:
-    'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span'.
+  <Prop
+    name="headingAs"
+    type="'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span'"
+    default="h3"
+  >
+    The heading tag used for the title. It should be one of the following
+    strings: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span'.
   </Prop>
   <Prop name="actionItem" type="React.ComponentType">
-    An action item, most commonly a button or a link, related to the title.
-    The actionItem is <b>hidden for extra small viewports</b> (less than 480px).
+    An action item, most commonly a button or a link, related to the title. The
+    actionItem is <b>hidden for extra small viewports</b> (less than 480px).
   </Prop>
   <Prop name="hideActionItemOn" type="MQPartial<boolean>" default="{xs: true}">
     An object to specify in which breakpoints to hide the action item.
   </Prop>
   <Prop name="overrides" type="object">
-    If provided, overrides the respective presets for the component and provided elements.
+    If provided, overrides the respective presets for the component and provided
+    elements.
     <Prop name="stylePreset" type="MQ<string>">
       If provided, this overrides the style preset applied to the TitleBar.
     </Prop>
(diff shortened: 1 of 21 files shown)


src/title-bar/styled.tsx
```diff
@@ -6,10 +6,13 @@ import {
...
 import {ContainerProps} from './types';
+import {logicalPaddingProps} from '../utils/logical-properties';
 
 export const StyledStackContainer = styled(Stack)<ContainerProps>`
-  ${getResponsiveSpacingInset('titleBar', '')};
-  ${getStylePreset('titleBar')};
+  // LOGICAL_PROPS_TO_DO: remove the below func when logical props are used in defaults
+  ${getResponsiveSpacingInset('titleBar', '')}
+  ${getStylePreset('titleBar')}
+  ${logicalPaddingProps('titleBar')}
```
evgenitsn, line 15: I wasn't sure if I should call this function here or not. The same for the `TextField` component `styles.tsx`. 
Let me know if there is any issue with that.

src/text-field/styled.tsx
```diff
@@ -22,6 +23,9 @@ export const StyledInput = styled.input<
   ::placeholder {
     color: ${({placeholderColor}) => placeholderColor && placeholderColor};
   }
+  // LOGICAL_PROPS_TO_DO: remove the below func when logical props are used in defaults
   ${({$size}) =>
     getResponsiveSpace('padding', `textField.${$size}`, '', 'spaceInset')}
+
+  ${({$size}) => logicalPaddingProps(`textField.${$size}`)}
```
evgenitsn, line 30: Same here

newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-1986-layout-props-6 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-1986-layout-props-6/storybook

src/title-bar/__tests__/__snapshots__/title-bar.test.tsx.snap
```diff
@@ -925,6 +925,152 @@ exports[`TitleBar Should display correctly with action item as: Link 1`] = `
...
+  -webkit-box-pack: justify;
+  -webkit-justify-content: space-between;
+  justify-content: space-between;
+  margin-inline: 48px;
+  margin-block: 32px;
+  padding-inline: 8px;
+  padding-block: 16px;
+  background-color: transparent;
+  padding-inline: 8px;
+  padding-block: 16px;
```
mstuartf, line 952: The padding logical props are duplicated here. I think this is because you are calling `logicalPaddingProps` in the TitleBar styled component but both padding and margin logical props are already applied by the Stack (which the styled component inherits from).

src/title-bar/styled.tsx
mstuartf: I think it is redundant (see my comment on the snap file).

src/text-field/styled.tsx
mstuartf: I think this is also unnecessary because the logical props are already applied to the root element (`StyledInputContainer` via `WithEnhancers`). This call would apply the same padding logical props again to the `StyledInput`.

src/tabs/__tests__/__snapshots__/tabs.test.tsx.snap
```diff
@@ -30375,3 +30375,708 @@ exports[`tabs indicatorPosition renders vertical tabs with the selection indicat
...
+  -ms-flex-direction: row;
+  flex-direction: row;
+  -webkit-box-pack: start;
+  -ms-flex-pack: start;
+  -webkit-justify-content: flex-start;
+  justify-content: flex-start;
+  margin-inline: 48px;
+  margin-block: 32px;
+  padding-inline: 8px;
+  padding-block: 16px;
```
mstuartf, line 30443: Here the logical props are being applied to `StyledInnerTabGroup` (using the `Stack` logic). Is this correct? Visually, this means that the logical props will not be applied to the tabs content, only the clickable tabs at the top. 

Screenshot 2022-05-27 at 15 48 40

If we want the logical props to apply to the root element, then `StyledTabGroup` needs a `logicalProps` call and the logical props need to be stripped out of the overrides passed to all subcomponents using `omitLogicalPropsFromOverrides`.

site/pages/components/title-bar.mdx
```diff
@@ -45,15 +51,23 @@ Title bar has a number of props to facilitate a variety of uses:
     </Prop>
```
mstuartf, line 51: This prop has been deprecated.

mstuartf: Same for `TextField` I think.

src/tabs/__tests__/__snapshots__/tabs.test.tsx.snap
evgenitsn: Addressed all comments 🚀

mutebg (approved):

PR #217, opened 2022-05-31
feat(PPDSC-2118): tooltip pointer

mstuartf:
[PPDSC-2118](https://nidigitalsolutions.jira.com/browse/PPDSC-2118)

**What**

1. Background - v1 tooltip did not have a pointer arrow or ability to adjust distance
2. What did you do - added optional pointer element to tooltip and ability to control distance

**I have done:**

**I have tested manually:**

**Differences to ticket:**
* Did not allow for setting pointer position separately. The library that we are using doesn’t really allow for this. The position of the pointer is tied to the position of the tooltip panel (e.g. image if the panel were left and the pointer right). Unclear what the use case would be. [Confirmed with Mike M]
* The ticket has `distance` as prop and an override. I only included it as an override as this seems more appropriate. 
* Pointer should just inherit background colour from the main tooltip. [Confirmed with Mike M]

**Other notes:**
* Changed the tooltip element type to be a `<div>` rather than a `<p>`. This is because the pointer `<div>` needs to be a child of the tooltip so that it can be positioned correctly. But `<div>` cannot appear as a descendant of `<p>` (or we get a `validateDOMNesting` console error). 
* Updated default trigger for the tooltip to be ['hover', 'focus'], rather than just 'hover'. This is so that tooltips work with screen readers by default. [Confirmed with Xin]
* Nested panel overrides in line with ticket spec. [Confirmed with Xin and Mike M]


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2118-tooltip-pointer 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2118-tooltip-pointer/storybook

src/tooltip/defaults.ts
```diff
@@ -1,9 +1,16 @@
...
     zIndex: 80,
-    paddingBlock: 'spaceInset020',
-    paddingInline: 'spaceInset020',
     stylePreset: 'tooltip',
+    distance: 'space020',
+  },
+  tooltipPointer: {
+    size: 'sizing010',
+  },
+  tooltipPanel: {
```
mutebg, line 10: I believe we should use nested objects
```
export default {
  tooltip: {
    zIndex: 80,
    stylePreset: 'tooltip',
    distance: 'space020',
    pointer: {
      size: 'sizing010',
    },
    panel: {
      stylePreset: 'tooltipPanel',
      typographyPreset: 'utilityLabel010',
      paddingBlock: 'spaceInset020',
      paddingInline: 'spaceInset020',
    },
  },
};
```

src/tooltip/styled.tsx
```diff
@@ -15,6 +16,42 @@ export const StyledTooltip = styled(TextBlock)<Pick<TooltipProps, 'overrides'>>`
   ${getResponsiveSize('maxWidth', 'tooltip', '', 'maxWidth')};
   ${getResponsiveSize('minWidth', 'tooltip', '', 'minWidth')};
   ${getStylePreset('tooltip', '')};
-  ${getTypographyPreset('tooltip', '')};
-  ${logicalProps('tooltip', '')}
+  ${getStylePreset('tooltipPanel', 'panel')};
```
mutebg, line 19: getStylePreset is used 2 times with different paths
also if that's the panel can we rename the component to StyledPanel

src/tooltip/tooltip.tsx
```diff
@@ -8,36 +8,66 @@ import {
...
-import {StyledTooltip} from './styled';
+import {StyledPointer, StyledTooltip} from './styled';
 import defaults from './defaults';
 import stylePresets from './style-presets';
 import {useControlled} from '../utils/hooks';
+import {useTheme} from '../theme';
+import {getResponsiveSpace, ThemeProp} from '../utils';
+
+const calculateDistance = <Props extends ThemeProp>(props: Props) => {
+  const {distance} = getResponsiveSpace(
```
mutebg, line 26: according to the types, this can be MQ<string>, if you want to use this function better change the type

src/tooltip/tooltip.tsx
```diff
@@ -8,36 +8,66 @@ import {
...
+    'distance',
+    'tooltip',
+    'distance',
+    'distance',
+  )(props) as {distance: string};
+  return parseInt(distance.replace('px', ''), 10);
```
mutebg, line 32: what if the value is not `px`?

mutebg (changes requested):

src/tooltip/tooltip.tsx
mstuartf: Oh yes, good point. 

Is it better to change the type or to handle MQs and follow the spec? It seems likely that users might want to set distance conditional on screen size. 

If we do want to handle MQs, we'd need to calculate the offset inside the `getResponsiveSpace` callback using logic like this: https://github.com/floating-ui/floating-ui/blob/master/packages/core/src/middleware/offset.ts#L32, then update the `left` and `top` values for the div. Annoyingly I don't think we can just use this function from the lib because it is async. It's not too complicated but it would mean duplicating some of the library logic on our side, which I don't love. What do you think?

src/tooltip/styled.tsx
mstuartf: It's called twice because the panel has its own styling (border radius, color) and the tooltip styling that it shares with the pointer (background color). Is there a better way to do this?

mutebg: I understand why is like that but still seems weird to me, there can be conflicts, like if both have borders? the panel ones will always override the tooltip ones.  Maybe we can do something like this ( as the designer envisioned hat )
```
<StyledTooltip>
  <StyledPanel />
  <StyledPointer />
</StyledTooltip>
```
image

src/tooltip/tooltip.tsx
mutebg: What do you think about setting an offset via padding on the tooltip container? something like this https://codepen.io/mutebg/pen/JjpBwqv that case we don't use floating UI functions we do it by ourselves, which will allow also use non-px values. I am not sure if that's gonna work just had that idea

src/tooltip/styled.tsx
mstuartf: I agree that having a containing tooltip element makes sense. Happy to implement that.

src/tooltip/tooltip.tsx
mstuartf: We'd still need to duplicate the library logic (to know which to side to apply the padding). But I don't think there's any way round that.

mutebg: my idea is not to use the offset function at all, maybe we can catch up on that tomorrow, I might need to better understanding of how it works

src/tooltip/styled.tsx
mstuartf: Actually this will not solve the problem. If the background color is applied to the tooltip element, the pointer (which is positioned absolutely to sit outside the container) does not inherit this. Also the border radius of the panel would need to be applied to the tooltip element, rather than to the panel.

The options are:
* Call `getStylePreset` twice to apply shared and specific styling.
* Set background color twice (once for the panel and once for the pointer. 

(But I do still think it makes sense to have a separate container element.)

src/tooltip/tooltip.tsx
mstuartf: Yeah, I just mean that we will be re-implementing some logic that the library already handles. But I think that it's unavoidable.

mstuartf: Screenshot 2022-06-06 at 14 48 11

Using padding doesn't work with how the lib implements the points (relative to the tooltip element).

src/tooltip/tooltip.tsx
mstuartf: Good point. Handled using `calc`.

src/tooltip/styled.tsx
mstuartf: For now I've reverted to setting background color separately for the panel and the pointer. Let me know what you think.

src/tooltip/tooltip.tsx
mstuartf: For now I'm updating the `top` / `left` values in the styled component. This supports non-px values and MQs. Let me know what you think.

mutebg: IMO looks very good.

src/tooltip/styled.tsx
mutebg: I personally like it this way

src/tooltip/__tests__/__snapshots__/tooltip.test.tsx.snap
```diff
@@ -1,20 +1,161 @@
...
+  bottom: calc(-4px / 2);
+}
+
+<div
+    class="Tooltip emotion-0"
+    id="floating-ui-38"
+    role="tooltip"
+    tabindex="-1"
+    x="0"
+    y="0"
```
mutebg, line 57: can you remove these attributes from the HTML, you can rename your prop to $x and $y so that they don't end up in the HTML or use [shouldForwardProp](https://emotion.sh/docs/styled#customizing-prop-forwarding)

src/tooltip/tooltip.tsx
```diff
@@ -21,23 +23,34 @@ const ThemelessTooltip: React.FC<TooltipProps> = ({
   children,
   title,
```
mutebg, line 24: I know this is not from your ticket but do you know why is called `title`, isn't it better to be named `content` or `render`? 
@Xin00163

mutebg (commented):
just 1 small comment, well done 👏

src/tooltip/tooltip.tsx
Xin00163: This is the name Mike had in design. I didn't feel too strongly about it but if you all think `content` is better, we can change it.

mstuartf: Would this count as a breaking change if the first version of the tooltip has been released? Same for the changes to overrides actually.

mstuartf: OK cool, done. All changes made.

mutebg (dismissed):

Xin00163 (commented):
Could we add the `aria-hidden` attribute to tooltip like https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA1? This is something I missed in the last pr. When scanning through with screenreader, the content of the tooltip is read twice. By adding this, we can avoid redundancy. Happy to have a chat about this.

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -65,24 +81,136 @@ describe('Tooltip', () => {
...
           },
         },
         myCustomTheme,
       );
       expect(asFragment()).toMatchSnapshot();
     });
+    test('with pointer', async () => {
+      const {asFragment} = await asyncRender(Tooltip, {
+        ...defaultProps,
+        showPointer: true,
```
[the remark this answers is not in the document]
mstuartf: Yes, that makes sense.

src/tooltip/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function getOffsetAxis(placement: Placement): Axis {
+  return !['top', 'bottom'].includes(getSide(placement)) ? 'x' : 'y';
+}
+
+export function getOffsetAxisDirection(placement: Placement): 1 | -1 {
+  return ['left', 'top'].includes(getSide(placement)) ? -1 : 1;
+}
+
+export function calculateInset(
```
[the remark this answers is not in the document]
mstuartf: 👍

mutebg (approved):

Xin00163 (approved):

PR #221, opened 2022-06-02
fix(PPDSC-2203): handling non text children in Paragraph

mutebg:
PPDSC-2203

**What**

1. Background - drop-cap on the paragraph component is broken when you use no-text only children like `text <b>other</b>` since for React treads that as multiple children
2. What did you do - 
    2.1. - Updated code so that handles multiple children, and  adjusted the ScreenOnly content, to wrap only first word instead of the whole paragraph text
    2.2. - When the first child is not a text it will not show the drop-cap.
    2.3. - Added support for React.Fragment

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/typography/paragraph/paragraph.tsx
+++ b/src/typography/paragraph/paragraph.tsx
@@ -1,18 +1,18 @@
 import React from 'react';
+import {isFragment} from 'react-is';
 import {
   styled,
   getTypographyPreset,
   MQ,
-  getSpace,
   getStylePreset,
+  getResponsiveSpace,
 } from '../../utils/style';
 import defaults from './defaults';
 import {withOwnTheme} from '../../utils/with-own-theme';
 import {ScreenReaderOnly} from '../../screen-reader-only';
 
 export interface ParagraphProps {
-  // eslint-disable-next-line
-  children: any;
+  children: React.ReactNode;
   dropCap?: boolean;
   overrides?: {
     stylePreset?: MQ<string>;
@@ -35,11 +35,9 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
 });
 
 const ThemelessParagraphDropCap = styled.span<ParagraphProps>`
-  margin: 0;
+  margin: 0 0 0 0.15em;
   float: left;
-  margin-right: 0.15em;
-
-  margin-top: ${getSpace('paragraph.dropCap', 'dropCap')};
+  ${getResponsiveSpace('marginTop', 'paragraph.dropCap', 'dropCap', 'space')};
   ${getTypographyPreset('paragraph.dropCap', 'dropCap')};
   ${getStylePreset('paragraph.dropCap', 'dropCap')};
 `;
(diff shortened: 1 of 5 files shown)


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2203-paragraph-dropcap 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2203-paragraph-dropcap/storybook

Srinivasan-Ramasamy (dismissed):

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,86 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
+  if (children.length > 0 && typeof children[0] === 'string') {
+    return children[0].split(' ')[0];
+  }
+  if (children.length && isFragment(children[0])) {
+    return getFirstWord(React.Children.toArray(children[0].props.children));
+  }
+  return '';
+};
+
+const removeFirstWord = (
```
mstuartf, line 64: What do you think about making this a pure function that returns a new array rather than modifying the input?

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,86 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
 });
 
-const ParagraphContainer = styled.div`
-  display: inline-flex;
-  max-width: 100%;
-`;
+const getFirstWord = (
+  children: (React.ReactChild | React.ReactFragment | React.ReactPortal)[],
+): string => {
+  if (children.length > 0 && typeof children[0] === 'string') {
```
mstuartf, line 55: The two `if` conditions for `getFirstWord` and `removeFirstWord` are identical. Could be abstracted?

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,86 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
 `;
 export const ParagraphDropCap = withOwnTheme(ThemelessParagraphDropCap)({
   defaults,
 });
 
-const ParagraphContainer = styled.div`
-  display: inline-flex;
-  max-width: 100%;
-`;
+const getFirstWord = (
```
Vanals, line 52: Should this be "getFirstLetter" ?

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,86 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
+<p>
+  <span aria-hidden="true"><span class="dropcap">T</span>his</span>
+  <span class="sr-only">This</span> is paragraph text
+</p>          
+
+ */
 
 export const Paragraph: React.FC<ParagraphProps> = ({
-  children,
+  children: ch,
```
Vanals, line 92: I would keep "children", why are we using an alias such as `ch`?  I think makes it bit less easy to understand what is it

src/typography/paragraph/paragraph.tsx
mutebg: tried this approach:
```
const firstChildrenIsString = (children) => {
  return children.length > 0 && typeof children[0] === 'string')
}
```
and later
```
if ( firstChildrenIsString(children) ) {
    return children[0].split(' ')[0];
  }
  ```
  but TS does not like it, it complains that children[0] is not a string, and its the same for the fragment condition. I don't want to cast it si prefer to keep it that way. Do you have any ideas?

File: src/typography/paragraph/paragraph.tsx [hunk 5]
```diff
@@ -34,41 +34,88 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
+    <ParagraphText>
+      {useDropCap && (
+        <>
+          <span aria-hidden="true">
+            <ParagraphDropCap overrides={overrides}>
+              {firstWord[0]}
+            </ParagraphDropCap>
+            {firstWord.slice(1)}{' '}
+          </span>
+          <ScreenReaderOnly>{firstWord}</ScreenReaderOnly>
```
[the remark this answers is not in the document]
mutebg: I have changed it a bit, let me know what you think

src/typography/paragraph/paragraph.tsx
mutebg: thanks, I removed `children.length` conditions since they will be covered by `typeof firstChild === 'string'` and `isFragment(firstChild)`, before was using indexed array like children[0] and that's why needed them

Xin00163 (commented):
could we have a storybook scenario? The rest looks good

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,88 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
   defaults,
 });
 
+/*
+We use this solution instead of css :first-letter since there is Firefox inconsistences,
+which causes the first letter to has less margin than desired.
+*/
 const ThemelessParagraphDropCap = styled.span<ParagraphProps>`
-  margin: 0;
+  margin: 0 0.15em 0 0;
```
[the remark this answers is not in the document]
Xin00163: Actually it doesn't have any impact. Ignore me

mstuartf (approved):

mstuartf (approved):

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,72 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
-const ParagraphContainer = styled.div`
-  display: inline-flex;
-  max-width: 100%;
-`;
+const getFirstLetter = (
+  children: (React.ReactChild | React.ReactFragment | React.ReactPortal)[],
+): string => {
+  const [firstChild] = children;
+  if (typeof firstChild === 'string') {
+    return firstChild.split('')[0];
```
evgenitsn, line 57: Tiny but you can use `.charAt` for getting the first letter of a string like so
`return firstChild.charAt(0);`

src/typography/paragraph/paragraph.tsx
```diff
@@ -34,41 +34,73 @@ export const ParagraphText = withOwnTheme(ThemelessParagraphText)({
...
+  }
+  return '';
+};
+
+const removeFirstLetter = (
+  children: (React.ReactChild | React.ReactFragment | React.ReactPortal)[],
+): (React.ReactChild | React.ReactFragment | React.ReactPortal)[] => {
+  const [firstChild, ...rest] = children;
+  if (typeof firstChild === 'string') {
+    return [firstChild.split('').slice(1).join(''), ...rest];
```
evgenitsn, line 70: Similarly I think we can use `.substring(1)` instead of `firstChild.split('').slice(1).join('')`.

mutebg: good ones, thanks

Xin00163 (approved):

mstuartf (approved):

PR #230, opened 2022-06-13
fix(PPDSC-2199): docs-ppdse-2199-drawer-correct-state

baburay23:
PPDSC-2199

**What**

1. Background - When a button was toggled inside the drawer, it was loosing the React Transition Group className. Which meant the button toggle worked but the Drawer immediately closed. This was also happening on the Modal which uses React Transition Group
2. What did you do- Used  Switch statement to track the state of the className as it transitions and then changed the Classname according to the state. This worked and the Drawer/ Modal is now able to maintain the transition state
you open the drawer(`enter-active , enter-done`) transitions occur
 then you toggle the theme button - now the `enter-done` transition remains thanks to the Switch statement
then you can close the drawer and `exit-active exit-done` transition will play out

if you would like to try it yourself there is a storybook code example in the ticket you can copy 

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2199-drawer-correct-state 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2199-drawer-correct-state/storybook

src/drawer/utils.ts
```diff
@@ -47,3 +47,34 @@ export const setDrawerElementFocusability = (
...
+    return 'nk-modal';
+  }
+  return 'nk-drawer';
+};
+
+export const setCorrectClassName = (
+  componentName: string,
+  state: string,
+): string => {
+  let prefixClassName = '';
```
mutebg, line 62: technically this is a `suffix`

src/drawer/utils.ts
```diff
@@ -47,3 +47,34 @@ export const setDrawerElementFocusability = (
     });
   }
 };
+
+const whichComponent = (prefix: string) => {
+  if (prefix === 'nk-modal') {
+    return 'nk-modal';
+  }
+  return 'nk-drawer';
+};
```
mutebg, line 56: Not sure if we need to have this function, for now, we only have 2 components and its easy but every time we need to use the function we need to add another statement to that if/else.
I personally wound go with `setCorrectClassName(prefix, state) => ...` and use it like `setCorrectClassName('nk-drawer', state)`

src/modal/modal.tsx
```diff
@@ -12,6 +12,7 @@ import {getTransitionDuration} from '../utils/get-transition-duration';
 import defaults from './defaults';
 import stylePresets from './style-presets';
 import {withOwnTheme} from '../utils/with-own-theme';
+import {setCorrectClassName} from '../drawer/utils';
```
mutebg, line 15: maybe is better to move this function to `src/utils/` so it's more obvious that is a shared one.

src/modal/__tests__/modal.stories.tsx
```diff
@@ -47,7 +47,47 @@ export default {
   component: () => 'None',
   disabledRules: ['tabindex'], // Because of scenario 'open on page load'
 };
-
+// export const StoryDefaultThemeBug = () =>
+//   React.createElement(() => {
+//     const [isActive, setIsActive] = React.useState(false);
+//     const [themeMode, setThemeMode] = React.useState('light');
```
mutebg, line 53: just cleanup this one, or un-comment it so people can test it.

src/drawer/utils.ts
```diff
@@ -47,3 +47,34 @@ export const setDrawerElementFocusability = (
...
+
+export const setCorrectClassName = (
```
mutebg, line 58:
```suggestion
export const getTransitionClassName = (
```

mutebg (changes requested):

src/drawer/utils.ts
baburay23: my bad i went a bit over board!

src/modal/__tests__/modal.stories.tsx
baburay23: i have removed it i thought if people would like to test they can use the example in the tickeT?

mstuartf (changes requested):
If I'm understanding correctly, the bug was triggered when the Theme changed. Can we add a test that covers this case? I.e. a test that changes the Theme when a Drawer is open, and then asserts that the Drawer is still open?

baburay23:
> 



> If I'm understanding correctly, the bug was triggered when the Theme changed. Can we add a test that covers this case? I.e. a test that changes the Theme when a Drawer is open, and then asserts that the Drawer is still open?

@mstuartf yes thats the case. Thanks for pointing it out i have written a test that clicks on the toggle theme button inside the drawer and i make sure the drawer has the correct transition class added to it `nk-drawer-enter-done` this confirms the drawer remains open after the theme is toggled. without the className the drawer would close

mstuartf (dismissed):

src/drawer/__tests__/drawer.test.tsx
```diff
@@ -5,37 +5,74 @@ import {
...
+    const {findByTestId, getByTestId} = renderWithTheme(DrawerPage);
+    const openDrawerButton = getByTestId('drawer-open-button');
+    fireEvent.click(openDrawerButton);
+    await waitFor(async () => {
+      const toggleThemeButton = await findByTestId(
+        'toggle-theme-drawer-button',
+      );
+      fireEvent.click(toggleThemeButton);
+    });
+    expect(getByTestId('drawer')).toBeInTheDocument();
```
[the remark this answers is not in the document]
baburay23: yes you are right sorry, i will remove and just use `tohaveClass` as this tests its still open

mutebg (approved):

Srinivasan-Ramasamy (approved):

PR #234, opened 2022-06-21
feat(PPDSC-2156): accordion group

mutebg:
PPDSC-2156

**What**

1. Background - we need a way to control Accordion component
2. What did you do - Implemented AccordionGroup which can be used in a controlled or uncontrolled way:
2.1.  `defaultExpanded` prop: you can pass index(es)  of the initially expanded accordion, this is when the component is used as uncontrolled; passing 'all' as value will expand all accordions
2.2. `expanded` prop: you can pass index(es)  of the expanded accordion, this is when is `controlled`; passing 'all' as value will expand all accordions
2.3. `onChange ` prop: triggers every time accordion expands/collapses and emits expanded indexes
2.4. `expandSingle` prop: allow only a single accordion to be open at a time.

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2156-accordion-group 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2156-accordion-group/storybook

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,81 @@
...
+        setExpandedState(newExpandedState);
+
+        if (onChange) {
+          onChange(newExpandedState);
+        }
+      },
+      [setExpandedState, expandedList, single, onChange],
+    );
+
+    const accordionOnlyChildren = React.Children.toArray(
```
mstuartf, line 64: Would it be better to show a warning, rather than silently filtering out the invalid children?

src/accordion/types.ts
```diff
@@ -36,4 +36,15 @@ export interface AccordionProps {
   expanded?: boolean;
   onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
   overrides?: AccordionPropsOverrides;
+  onChange?: (expanded: boolean) => void;
 }
+
+export type AccordionGroupProps = {
+  defaultExpanded?: number | number[];
```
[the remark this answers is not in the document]
mutebg: I am not sure, since I don't think we saved what we discussed during TA. I saw other libraries like Radix and Ant are supporting this

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }) as [number[], (state: number[]) => void];
+
+    const handleChange = useCallback(
+      (isExpanded: boolean, index: number) => {
+        let newExpandedState: number[] = [];
+
+        // in the case only 1 accordion is allowed to be open
+        if (single) {
+          // replace current the expanded accordion with the new one
+          if (isExpanded && !expandedList.includes(index)) {
```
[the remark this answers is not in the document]
mutebg: That's correct, will update it.

src/accordion/accordion-group.tsx
mutebg: decided to remove that filter and allow any kind of children, and later cloning only accordion components

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,120 @@
...
+      children,
+      defaultExpanded = [],
+      onChange,
+      expanded,
+      expandSingle,
+      ...restProps
+    },
+    ref,
+  ) => {
+    const accordionOnlyChildren = React.Children.toArray(children);
```
mstuartf, line 53: I think this should just be `children` now?

src/accordion/types.ts
```diff
@@ -36,4 +36,15 @@ export interface AccordionProps {
...
 }
+
+export type AccordionGroupProps = {
+  defaultExpanded?: number | number[] | 'all';
+  expanded?: number | number[] | 'all';
+  onChange?: (expanded: number[]) => void;
+  expandSingle?: boolean;
+  children: React.ReactNode;
+  // | React.ReactElement<AccordionProps>
+  // | React.ReactElement<AccordionProps>[];
```
mstuartf, line 49: I think this should be removed

mstuartf (commented):
Two small changes otherwise looks good 👍

mstuartf (dismissed):

baburay23 (commented):
looks good to me

baburay23 (dismissed):

Srinivasan-Ramasamy:
Nice work @mutebg  :+1:

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,116 @@
+import React, {useCallback} from 'react';
+import {useControlled} from '../utils/hooks';
+import {AccordionGroupProps, AccordionProps} from './types';
+import {hasMatchingDisplayNameWith} from '../utils/component';
+import {Accordion} from './accordion';
+
+const toArray = (
+  value: number | number[] | 'all' | undefined,
```
Vanals, line 8: is the "value" actually the accordionIndex ? could we rename it as such to be clear?

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,116 @@
...
+    return Array.from(Array(arrayLength).keys());
+  }
+
+  if (value !== undefined && !Array.isArray(value)) {
+    return [value];
+  }
+  return value;
+};
+
+const maybeSingle = (
```
Vanals, line 21: what maybeSingle means 👀 ?

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,116 @@
...
+  array: number[] | undefined,
```
Vanals, line 22: can we rename this one specifying what list this function will receive? is the list of accordionIndexes?

src/accordion/accordion-group.tsx
```diff
@@ -0,0 +1,116 @@
...
+  arrayLength: number,
+): number[] | undefined => {
+  if (value === 'all') {
+    return Array.from(Array(arrayLength).keys());
+  }
+
+  if (value !== undefined && !Array.isArray(value)) {
+    return [value];
+  }
+  return value;
```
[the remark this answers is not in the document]
mutebg: the `expanded` prop can take `'all' | number | number[]` the idea of that function is to unify its value and return an array

src/accordion/accordion-group.tsx
mutebg: its not just an index, it can be 'all' string

src/accordion/accordion-group.tsx
mutebg: renamed to `expandedProp`

src/accordion/accordion-group.tsx
mutebg: rename to `takeOnlyFirst`

src/accordion/accordion-group.tsx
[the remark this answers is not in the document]
Vanals: Yes! what I thought was... if `expanded` is an array, you don't really need to pass it to `toArray`  right? cos it is already an array? and the logic calling `toArray` will use instead directly `expanded` rather of calling `toArray` if not needed.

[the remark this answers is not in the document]
mutebg: the logic is to unify expanded prop value, I renamed it to `expandedPropToArray`

Vanals:
I've pulled the changes but I can't see the accordion group scenarios
Screenshot 2022-06-24 at 09 16 27

baburay23 (approved):

PR #235, opened 2022-06-21
feat(PPDSC-2180): theme switch added to storybook

tbradbury:
PPDSC-2180

**What**

1. So we can see what components look like on different themes.
2. Extended the background addon to switch between themes. You can switch from Newskit light to Dark. I've also added in The Sun, Times and Virgin
3. changing themes changes components look.

Screenshot 2022-06-22 at 10 52 31 am

Screenshot 2022-06-22 at 10 52 46 am

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


mstuartf (commented):
Why are the changes to the AudioPlayer and Theme tests required?

Vanals:
Can you please fill in the check list?

Vanals:
Has this been reviewed also by designers?

I can see some of the components/scenarios do not look good with the new theme. I did not went through all of them, but for example some of the Card ones:

Screenshot 2022-06-22 at 13 47 18

Screenshot 2022-06-22 at 13 47 35

src/toast/__tests__/toast.stories.tsx
```diff
@@ -24,9 +26,10 @@ const CustomToast = styled.div`
   color: red;
 `;
 
-const myCustomTheme = compileTheme(
+const getCustomTheme = (theme: UncompiledTheme): UncompiledTheme =>
```
mutebg, line 29: we have this getCustomTheme on many places, wondering if we can create a function createStorybookTheme which makes things easier, like knowing which is current theme and use it, instead of passing doing that on every single story

.storybook/preview.js
```diff
@@ -40,13 +41,33 @@ export const parameters = {
...
+  backgrounds: {
+    default: 'Newskit light',
+    values: [
+      {
+        name: 'Newskit light',
+        value: '#ffffff',
+      },
+      {
+        name: 'Newskit dark',
+        value: '#2E2E2E',
```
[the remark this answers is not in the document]
tbradbury: I'm using the background addon that comes with Storybook. The value changes the background of the story page, and we are also watching when this happens and changing themes accordingly. We don't want to change the background of the other themes they all need to be white, but we still need to change the theme. Not putting it a hex colour achieves this.

.storybook/preview.js
```diff
@@ -79,7 +100,7 @@ export const decorators = [
   },
   (Story, context) => {
     return (
-      <ThemeProvider theme={getThemeFromContext(context)}>
+      <ThemeProvider theme={themeObject[context?.globals?.backgrounds?.value || '#ffffff']}>
```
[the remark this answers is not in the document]
tbradbury: Will take a look

.storybook/preview.js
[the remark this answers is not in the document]
mutebg: Do we need to do that? can we just have a container that wraps everything and is having `interfaceBackground`or something like that? In that case, we depend on the theme colour not on the plugin value

src/toast/__tests__/toast.stories.tsx
mutebg: Lets have a chat about that today

.storybook/preview.js
```diff
@@ -22,12 +24,30 @@ const unlimitedScenarios = [
...
+
 const Container = styled.div`
   max-width: 1024px;
   max-height: 768px;
   overflow: hidden;
 `;
-const LimitSizeDecorator = ({children}) => <Container>{children}</Container>;
+const Background = ({children}) => <BackgroundColor><PaddingReset>{children}</PaddingReset></BackgroundColor>
+const LimitSizeDecorator = ({children}) => <Background><Container>{children}</Container></Background>;
+const MediaQueryProviderDecorator = ({children}) => <Background><MediaQueryProvider>{children}</MediaQueryProvider></Background>
```
mutebg, line 50: I think, MediaQueryProviderDecorator should not need the background.

.storybook/preview.js
```diff
@@ -41,13 +61,32 @@ export const parameters = {
   viewport: {
     viewports: INITIAL_VIEWPORTS,
   },
+  backgrounds: {
+    values: [
+      {
+        name: 'Newskit light',
+        value: '#ffffff',
```
mutebg, line 68: if we don't use these values for a background color maybe we can put something more meaningfull like newskit-light, newskit-dark

src/toast/__tests__/toast.stories.tsx
tbradbury: I've change this to a function, but yes getting the current selected theme as the base is the whole point. Without it you only get the values of your custom theme. When you switch over to another theme say light to dark, if you don't have the correct base theme them none of the text switches colour from black to white.

.storybook/preview.js
tbradbury: The card component does not look correct without this on reloads (dark theme)

.storybook/preview.js
tbradbury: Using the hex code also add a colour ref to the background dropdown denoting the theme:

Screenshot 2022-07-01 at 2 10 56 pm

I think it looks better like this.

e2e/tests.tsx
```diff
@@ -57,6 +58,18 @@ export default function showTestcase() {
...
+    return (
+      <>
+        {Object.values(story)
+          .filter(storyComponent => typeof storyComponent === 'function')
+          .map(Story => (
+            <>{story.default.decorators[0](Story)}</>
+          ))}
+      </>
+    );
+  }
```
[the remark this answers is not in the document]
tbradbury: Will add a comment. It's checking to see if the story has a decorator and if so wrapping the Story component it it. This ensures the custom themes are applied to the components in the e2e tests.

src/theme/__tests__/__snapshots__/theme.test.ts.snap
```diff
@@ -95,11 +95,11 @@ Object {
...
-    "interactiveInverse010": "{{colors.whiteTint010}}",
-    "interactiveInverse020": "{{colors.whiteTint020}}",
-    "interactiveInverse030": "{{colors.white}}",
-    "interactiveInverse040": "{{colors.whiteTint070}}",
-    "interactiveInverse050": "{{colors.whiteTint080}}",
+    "interactiveInverse010": "{{colors.blackTint010}}",
+    "interactiveInverse020": "{{colors.blackTint020}}",
+    "interactiveInverse030": "{{colors.black}}",
+    "interactiveInverse040": "{{colors.blackTint070}}",
+    "interactiveInverse050": "{{colors.blackTint080}}",
```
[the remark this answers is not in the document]
tbradbury: This adds missing values to the dark theme. I would not say its a breaking change personally?

mutebg (approved):
thanks for that huge effort again

mutebg (approved):

LukeFinch:
Is the intention to serve the fonts for each theme in this storybook also?

Currently changing themes only display correctly if the user has the fonts installed locally.

mutebg:
> Is the intention to serve the fonts for each theme in this storybook also?
> 
> Currently changing themes only display correctly if the user has the fonts installed locally.

This seems the cause of the issue, and right now is also present on prod and staging storybook. Probably we need a separate ticket for it
Screenshot 2022-07-05 at 13 26 30

Xin00163:
> > Is the intention to serve the fonts for each theme in this storybook also?
> > Currently changing themes only display correctly if the user has the fonts installed locally.
> 
> This seems the cause of the issue, and right now is also present on prod and staging storybook. Probably we need a separate ticket for it Screenshot 2022-07-05 at 13 26 30

We have this ticket: https://nidigitalsolutions.jira.com/browse/PPDSC-2037. It's in ready for to do.

mutebg (dismissed):

Xin00163 (commented):
Can we remove the theme-checker theme switch as a result?

src/video-player/style-presets.ts
```diff
@@ -5,10 +5,10 @@ export default {
     base: {
       backgroundColor: '{{colors.transparent}}',
       borderStyle: 'solid',
-      borderColor: '{{colors.inkInverse}}',
+      borderColor: '{{colors.white}}',
       borderWidth: '{{borders.borderWidth020}}',
       borderRadius: '{{borders.borderRadiusCircle}}',
-      color: '{{colors.inkInverse}}',
+      color: '{{colors.white}}',
```
Xin00163, line 11: Don't think we can do this. `inkInverse` is a theme color. It's `'{{colors.white}}'` in light thme but `'{{colors.black}}'` in dark-theme.

tbradbury: That has come from design - "the video title and icon button should be white in the dark theme" but will double check with Mike tomorrow

Xin00163: It's likely the token value in dark-theme is wrong. I will still use theme color as in a different theme, inkInverse might be another color

mutebg: IMO the right color is `inkBase` its dark colour on the light theme and light colour on the dark theme.

tbradbury: Have spoken to Mike and he has confirmed that this is correct. Should stay white on both light and dark theme. Currently looking at the colour tokens this is only possible using {{colors.white}}.

mutebg (approved):

PR #238, opened 2022-06-24
feat(PPDSC-1961): added initialTime prop and test

agagotowiec:
PPDSC-1961

**What**

Added initialTime prop to the audio-player-composable to allow customers to start the audio from a specific time. 
Added a new scenario in the storybook with initial props both for time and volume.

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/audio-player-composable/__tests__/audio-player-composable.test.tsx
+++ b/src/audio-player-composable/__tests__/audio-player-composable.test.tsx
@@ -77,6 +77,17 @@ const AudioPropsAndVolumeControlWithInitialVolumeCollapsed: AudioPlayerComposabl
   ),
 };
 
+const AudioPropsWithInitialTime: AudioPlayerComposableProps = {
+  initialTime: 50,
+  src: 'https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3',
+  children: (
+    <>
+      <AudioPlayerTimeDisplay />
+      <AudioPlayerSeekBar />
+    </>
+  ),
+};
+
 const AudioPropsAndVolumeControlVertical: AudioPlayerComposableProps = {
   src: 'https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3',
   initialVolume: 0.2,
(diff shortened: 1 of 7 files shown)


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-1961-initial-start-point-audioplayer 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-1961-initial-start-point-audioplayer/storybook

src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -71,9 +72,15 @@ export const AudioPlayerComposable = ({
   });
 
   useEffect(() => {
+    // On render onTimeUpdate will be fired and 50 will be set as a value for currentTime state.
+    // I can't set this one to the currentTime state directly as the audioElement time
+    // will still be 0, currentTime will be overridden to 0 and the audio will start from 0
+    if (audioRef && audioRef.current) {
+      audioRef.current.currentTime = initialTime;
+    }
     setCurrentTime(0);
```
mstuartf, line 81: Should `currentTime` still be set to 0 here if there is an `initialTime`? Is it possible to just call `setCurrentTime` with the `initialTime`?

src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -71,9 +72,15 @@ export const AudioPlayerComposable = ({
   });
 
   useEffect(() => {
+    // On render onTimeUpdate will be fired and 50 will be set as a value for currentTime state.
```
mstuartf, line 75: I think `50` is the value from the story.

src/audio-player-composable/__tests__/audio-player-composable.test.tsx
```diff
@@ -902,6 +913,29 @@ describe('Audio Player Composable', () => {
       expect(asFragment()).toMatchSnapshot();
     });
   });
+  describe('initialTime prop', () => {
+    it('should render correctly with initial time', () => {
```
mstuartf, line 917: I might be wrong but I don't think this test is actually testing the impact of the `initalTime` prop. If you comment it out of `AudioPropsWithInitialTime`, the test still passes. I think this is because you are setting the time manually in the `timeUpdate` call below. Similarly, if you comment out the changes to `audio-player-composable.tsx` and then re-run this test, it still passes. Happy to discuss.

src/audio-player-composable/audio-player-composable.tsx
Vanals: @agagotowiec  I think 50 was meant to be initialTime 🙏

src/audio-player-composable/audio-player-composable.tsx
Vanals: The logic behind is a bit of a knot.

- We can't set intialTime in the currentTime default state because otherwise we get an error from reactRange and the whole app crashed.

- We can't just set initialTime to setCurrentTime here, because it will only update the currentTime state but not the audio HTML element  time. And if the audio HTML element time is different from currentTime state, on `onTimeUpdate` the state will be change to what the audio element had, so 0

src/audio-player-composable/__tests__/audio-player-composable.test.tsx
Vanals: Yes, you are right 🤔 . I have removed the object we were passing to `fireEvent.timeUpdate`, so it takes the value from the prop. 👍  Looks working good

src/audio-player-composable/audio-player-composable.tsx
agagotowiec: updated 👍🏻

src/audio-player-composable/audio-player-composable.tsx
mstuartf: ok 👍

mstuartf (approved):

Srinivasan-Ramasamy (approved):
Looks good :+1:

PR #247, opened 2022-06-30
chore(PPDSC-2170): Follow-up ticket for Audio Player v2 - Volume Control

Xin00163:
PPDSC-2170

- update Volume control based on new spec

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2170-audioplayer-volume-followup 
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2170-audioplayer-volume-followup/storybook

tbradbury (dismissed):

src/audio-player-composable/components/volume-control/styled.ts
```diff
@@ -1,29 +1,39 @@
...
-}>`
-  ${({vertical}) =>
-    vertical &&
-    `position: absolute;
-     bottom: 100%;`};
+  ${({layout}) =>
+    getResponsiveSpace(
+      layout === 'vertical' ? 'rowGap' : 'columnGap',
+      `audioPlayerVolumeControl.${layout}.spaceBetween`,
+      `${layout}.spaceBetween`,
```
mstuartf, line 36: I think this override path does not match the type - `spaceBetween` is not nested inside `layout` in the overrides object. I couldn't override this locally until I changed this line to just `spaceBetween`. 

You could update this path here, or we could nest the overrides inside of layout keys to match the default object structure, which might be clearer.

src/audio-player-composable/components/volume-control/types.ts
```diff
@@ -1,17 +1,23 @@
...
+import {NewsKitIconProps} from '../../../icons';
 import {SliderOverrideProps} from '../../../slider';
 import {MQ} from '../../../utils';
+import {Override} from '../../../utils/overrides';
 
+export type MuteButtonIconProps = NewsKitIconProps &
+  Pick<MuteButtonProps, 'volume'>;
 export interface AudioPlayerVolumeControlOverridesProps {
+  stylePreset?: MQ<string>;
+  spaceBetween?: MQ<string>;
```
mstuartf, line 11: Related to my other comment, does it make sense to nest these under layout keys? Would that allow more flexibility for setting space, styling, etc for the vertical and horizontal versions separately? It might also be clearer because the overrides object structure would match the default object structure.

src/audio-player-composable/components/volume-control/styled.ts
mstuartf: E.g. I don't think the `spaceBetween` override in the overrides story is being applied.

src/audio-player-composable/components/volume-control/types.ts
Xin00163: Yeah I agree. That makes the overrides more clear as to what layout it's changing.

mstuartf: Hmm. But what if I want to set global overrides for the horizontal and vertical layouts separately? I.e. at theme level.

Xin00163: We can override it with component defaults like this: https://github.com/newscorp-ghfb/newskit/blob/main/site/theme/component-defaults.ts#L214. Is this what you mean?

mstuartf: Yeah. But I mean that if we don't nest under layout keys, users will only be able to set one global override that will be applied to both layouts. Is that right? I can't add global component overrides for horizontal and vertical separately in the current structure. I might be understanding though (it may also be a non-issue).

mstuartf (dismissed):

src/audio-player-composable/components/volume-control/styled.ts
```diff
@@ -1,29 +1,39 @@
...
-  vertical?: boolean;
-}>`
-  ${({vertical}) =>
-    vertical &&
-    `position: absolute;
-     bottom: 100%;`};
+  ${({layout}) =>
+    getResponsiveSpace(
+      layout === 'vertical' ? 'rowGap' : 'columnGap',
+      `audioPlayerVolumeControl.${layout}.spaceBetween`,
```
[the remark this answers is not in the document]
Xin00163: Design has updated this to `spaceBetween`

mutebg (approved):

baburay23 (approved):

PR #256, opened 2022-07-04
feat(PPDSC-2176): feat-ppdse-2176-added tooltip to themecheck

baburay23:
PPDSC-2176

**What**

2. What did you do

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/theme-checker/scenarios.tsx
+++ b/src/theme-checker/scenarios.tsx
@@ -14,7 +14,7 @@ import {
 import {Tag} from '../tag';
 import {LinkInline, LinkStandalone} from '../link';
 import {IconButton} from '../icon-button';
-import {Button} from '../button';
+import {Button, ButtonSize} from '../button';
 import {
   IconFilledAddCircle,
   IconFilledEmail,
@@ -99,6 +99,7 @@ import {calculateTime} from '../audio-player-composable/components/time-display/
 
 import {GridLayout} from '../grid-layout';
 import {SwitchState} from '../switch/types';
+import {Tooltip} from '../tooltip';
 
 export const checkboxStates: [
   string,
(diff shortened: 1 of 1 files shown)


src/tooltip/__tests__/tooltip.stories.tsx
```diff
@@ -109,6 +125,7 @@ export const StoryTooltip = () => (
     >
       <Button
         size={ButtonSize.Small}
+        // disabled
```
Xin00163, line 128: comment?

Xin00163 (dismissed):

src/tooltip/tooltip.tsx
```diff
@@ -32,6 +32,18 @@ const ThemelessTooltip: React.FC<TooltipProps> = ({
...
+    if (process.env.NODE_ENV !== 'production' && children.props.disabled) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `When passing a component with disabled prop to Tooltip please remember to wrap it inside span element.`,
+      );
+    }
+  };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => showSpanWarning(), []);
```
mstuartf, line 45: Might be worth triggering this whenever `children.props.disabled` changes. The reference element may not be disabled on first render. Don't feel strongly though.

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -421,7 +430,19 @@ describe('Tooltip', () => {
...
     });
   });
-
+  test('should throw a warning if disabled component is not wrapped in a span', async () => {
+    jest.spyOn(console, 'warn').mockImplementation();
+    const {asFragment, getByRole} = renderWithTheme(Tooltip, {
+      ...disabledPropToolTip,
+    });
+    fireEvent.mouseEnter(getByRole('button'));
+    await applyAsyncStyling();
```
mstuartf, line 439: Is this (and the snapshot check below) necessary? The test is to check that the warning appears in the console.

src/tooltip/tooltip.tsx
baburay23: I was not really sure how to check is the props change. After some research I have added something inside useEffect to check  if children.props.disabled changes.
However i noticed when the page renders without the disabled prop and then i add it in without using  <span>  the correct error shows immediately . This happens on the old useEffect implementation as well  `useEffect(() => showSpanWarning(), []);`

src/tooltip/__tests__/tooltip.test.tsx
baburay23: ah sorry yes you are right i have removed it

src/tooltip/tooltip.tsx
```diff
@@ -32,6 +32,17 @@ const ThemelessTooltip: React.FC<TooltipProps> = ({
 
   const contentIsString = typeof content === 'string';
 
+  const showSpanWarning = (): void => {
```
mutebg, line 35:
```suggestion
  const showDisabledWarning = (): void => {
```

mstuartf (dismissed):

mutebg (approved):

Xin00163 (approved):

PR #264, opened 2022-07-06
feat(PPDSC-2115): accordion motion

mstuartf:
[PPDSC-2115](https://nidigitalsolutions.jira.com/browse/PPDSC-2155)

**What**

1. Background:
- Accordion v1 does not support transitions
3. What did you do:
For the header:
- Added support for transitions and set `backgroundColorChange` as default.

For the panel:
- Considered using `slideTop` (similar to Drawer component) but this does not work. The Drawer sits above existing content, but the Accordion panel needs to sit within the other elements on the page when expanded. 
- Considered using `scaleY` but this distorts the panel contents during the transition. 
- Considered using height but this is not possible when expanded height is unknown.
- Settled on using `maxHeight`, which seems to be the consensus: https://www.w3schools.com/howto/howto_js_accordion.asp
- Added a new `maxHeightChange` transition preset.
- Added support for transitions and set `maxHeightChange` as default.
- Conditionally set the panel max height based on `expanded` state. This needs to be set manually in the component as it is specific to each case (i.e. not a generic value like 100%). 

5. What does the reviewers should expect:
- Default transition presets working for accordion + ability to override

**I have done:**

**I have tested manually:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2115-accordion-motion-height
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2115-accordion-motion-height/storybook

src/accordion/accordion.tsx
```diff
@@ -16,6 +17,31 @@ import {getComponentOverrides} from '../utils/overrides';
...
+  return (
+    <StyledPanelTransitionContainer
+      ref={ref}
+      expanded={expanded}
+      overrides={overrides}
+    >
+      {children}
+    </StyledPanelTransitionContainer>
+  );
+};
```
mutebg, line 43: This seems good solution but I don't think is very flexible and it will not allow people to use other transition presets
You might need to use CSSTransition to make it more flexible.

mutebg (dismissed):

mutebg (dismissed):

Vanals:
Is this "Ready for review" ? I cannot see any ticket related to 2115 in the board or review column. If not can we remove the tag?

mstuartf:
> Is this "Ready for review" ? I cannot see any ticket related to 2115 in the board or review column. If not can we remove the tag?

It is flagged (in the To Do column). I'm still waiting to hear back from Applitools about the failing tests.

Vanals:
> > Is this "Ready for review" ? I cannot see any ticket related to 2115 in the board or review column. If not can we remove the tag?
> 
> It is flagged (in the To Do column). I'm still waiting to hear back from Applitools about the failing tests.

I see. I think is a bit confusing in this way?
A PR is either ready for review or is not. If it is should be in the ready to review Jira column(can flag it if blocked) and if not ready should not have the tag in github :D.

mutebg (approved):

baburay23 (approved):

PR #371, opened 2022-09-12
feat(PPDSE-2360): feat-2360-vertical-volume-control-gap

baburay23:
PPDSC-2360

**What**

Please note that this was originally a spike but no investigation was needed in the end, design were happy with the change so i implemented everything here in the ticket

1. Background - There is no space between vertical volume control popover and the mute button
2. What did you do- 
After some attempts we decided to extend the props of the Popover and allow the user to set `hidePointer=false` on the AudioPlayerVolumeControl. 
By bringing in the Popover Props and Overrides into the `AudioPlayerVolumeControl ` in audioPlayerVolumeControl the user can override the Pointer size and stylepreset, when the pointer is shown the user can hover between the mute button and popover, this is the only way to have the desired behaviour. We decided to give the option to the user to make the pointer transparent that way they can hover inbetween the gap.
If they create a style preset
```
const volumeTheme = createTheme({
  name: 'my-custom-volume-theme',
  overrides: {
    stylePresets: {
      customPointerStylePreset: {
        base: {
          backgroundColor: 'transparent',
        },
      },
    },
  },
});
```
and then add it to the pointer stylePreset override
```
        <AudioPlayerVolumeControl
          layout="vertical"
          overrides={{
            popover: {
              distance: 'space050',
              pointer: {
                size: 'sizing080',
                stylePreset: 'customPointerStylePreset',
              },
            },
          }}
        />
```
the pointer will now be transparent and the user can hoover
If they only passed `layout=vertical ` it would look like this
Screenshot 2022-09-16 at 12 52 32

Originally i thought it was a good idea to add the stylePreset for the pointer on our volumeControl defaults but after speaking to other engineers it seems quite hacky and this situation is an edge case. So we thought it might be better to add a storybook scenario instead and the user can set the transparency on the stylePreset themselves

I hope this is ok..if anyone has any opinions otherwise please do let me know.

4. What does the reviewers should expect- overridable space between mute button and popover on volume control vertical

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdse-2360-vertical-volumecontrol-gap
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdse-2360-vertical-volumecontrol-gap/storybook

mutebg (commented):
I think you are on the right path,
However, there is 1 issue. adding this styled block with extra padding makes the whole Block bigger and takes much more space than just a button.
IMO that should not happen.
Screenshot 2022-09-13 at 9 12 33

Xin00163 (approved):

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
```diff
@@ -1237,24 +1293,41 @@ AudioPlayerKeyboard.storyName = 'audio-keyboard-shortcuts';
...
+            popover: {
+              distance: 'space050',
+              pointer: {
+                size: 'sizing080',
+                stylePreset: 'customPointerStylePreset',
+              },
+            },
+          }}
+        />
+      </AudioPlayerComposable>
```
mutebg, line 1329: All the examples in the storybook are using the overrides, shall we show how this looks without overrides?
It might be confusing for some people.


If they import the component it will look like this, right?

Lets have  with and without overrides example

Screenshot 2022-09-16 at 16 58 28

src/audio-player-composable/components/volume-control/volume-control.tsx
```diff
@@ -129,9 +130,30 @@ const ThemelessAudioPlayerVolumeControl = React.forwardRef<
       overrides={sliderOverrides}
     />
   );
-
   const popoverOverrides = getPopoverOverrides(theme, overrides);
 
+  const TheContent = (
```
Vanals, line 135: Might this be renamed to `popoverContent` or `verticalVolumeSlider`?

src/audio-player-composable/__tests__/audio-player-composable.stories.tsx
mutebg: Also, since you are using the getComponentOverrides function in your types.ts you need to change to:

```
popover?: Override<PopoverProps>;
```

src/audio-player-composable/components/volume-control/volume-control.tsx
mutebg: only react component should start with the upper latter

src/audio-player-composable/components/volume-control/defaults.ts
```diff
@@ -23,6 +23,7 @@ export default {
       stylePreset: 'iconButtonMinimalPrimary',
     },
     popover: {
+      distance: 'space030',
```
Vanals, line 26: Just curious, should not be using padding or margin key words here? distance I think we have never used it? is that ok 🤔 ?

mutebg: I remember speaking with Mike F about that, its not margin or padding. that's `distance` between the popover and the button, you still can add padding and margin to the popover/tooltip.

Now I am seeing that probably the right name was "offset" ( that's the name in floating-ui library ) but it's too late since is gonna be a breaking change.

Vanals: I see! something maybe to consider for a major? for keeping our props consistent 🤔

mutebg (approved):

src/audio-player-composable/components/volume-control/defaults.ts
baburay23: @Vanals will create a ticket for this so that next time we work on major tickets we can fix this!

baburay23: https://nidigitalsolutions.jira.com/browse/PPDSC-2428

src/audio-player-composable/components/volume-control/volume-control.tsx
```diff
@@ -129,9 +130,30 @@ const ThemelessAudioPlayerVolumeControl = React.forwardRef<
...
-
   const popoverOverrides = getPopoverOverrides(theme, overrides);
 
+  const popoverContent = (
+    <StyledVolumeSliderPopupContainer overrides={overrides}>
+      {sliderComponent}
+    </StyledVolumeSliderPopupContainer>
+  );
+  const DefaultPopover = Popover;
+  const [PopoverComponent, popoverProps] = getComponentOverrides(
```
mstuartf, line 141: What's the difference between this and just having the Popover component in the return statement as before?

mutebg: the idea is to be able to override Popover props ( not only styling ) for example to set hidePointer=false

mstuartf: ok, thanks

Xin00163 (approved):

PR #374, opened 2022-09-15
feat(PPDSC-2145): add more menu storybook scenarios

evgenitsn:
PPDSC-2145

**What**

Added 3 new scenarios into 2 new pages under `menu` section about:
- 2 tier level menu navigation
- 3 tier level menu navigation
- mobile accordion navigation with sub-sections
(Screenshots below)

There was another part of this ticket about creating a scenario for audio player with tooltips attached to the buttons but due to internal audio player limitation we can't implement that at the moment. The issue is that when we Play the audio track the tooltips are flickering because of a re-rendering. 
If we want to have this we need to add tooltip support for the audio player sub-components internally. 

**I have done:**

**I have tested manually:**

**After:**
image
image

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2145-menu-storybook-scenarios
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2145-menu-storybook-scenarios/storybook

baburay23 (approved):

mstuartf (commented):
Do the examples need to be interactive?

src/menu/__tests__/menu.stories.tsx
```diff
@@ -1296,6 +1362,358 @@ export const StoryMenuItemsOutlineOverrides = () => (
...
+                  href={href}
+                  overrides={menuItemOverrides}
+                  size="small"
+                >
+                  {item}
+                </MenuItem>
+              ))}
+            </>
+          )}
+
```
mutebg, line 1643: This implementation does not seem right to me, it's again going to the previous comment with the correct HTML structure.
Also when the menu is closed it needs to be hidden, so that all links are still in the DOM but not visible.

mutebg (changes requested):

evgenitsn:
After a discussion with Stoyan about this. The current stories implementation is not semantically correct because it's not actually a sub-menus (second, third level) but just two or three independent menus.
The correct way would be to have a single menu and then inside to have nested menus attached to the top-level menu items. 

We currently have 2/3 separate menus without any semantic connection between them. This would be confusing for the screen readers and the tab order.

With our current Menu components we don't have the ability to create such nested menus that are semantically correct. 
We have to first investigate how to do this, then to update the Menu internal component to support this. When that's done it would be possible for us to make these scenarios in a correct way.

I am closing this PR now and will use it as a reference for the spike that I just created https://nidigitalsolutions.jira.com/browse/PPDSC-2433

PR #381, opened 2022-09-20
feat(PPDSC-2403): Extend role and aria-haspopup on Popover

jannuk59:
[PPDSC-2403](https://nidigitalsolutions.jira.com/browse/PPDSC-2403)

- Extended the `role` and `aria-haspopup` on the popover component.
Example:
```
<Popover content="..." role="menu">
  <IconButton aria-haspopup="true" />
</Popover>
```
- By default the `role` and `aria-haspopup` is to `'dialog'`

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2403-extend-role-and-aria-haspopup-on-popover
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2403-extend-role-and-aria-haspopup-on-popover/storybook

src/popover/popover.tsx
```diff
@@ -122,7 +123,7 @@ const ThemelessPopover = React.forwardRef<HTMLDivElement, PopoverProps>(
         buildRefElAriaAttributes={buildContextAriaAttributes}
         buildFloatingElAriaAttributes={buildFloatingElementAriaAttributes}
         useInteractions={useInteractions}
-        role="dialog"
+        role={role}
```
mutebg, line 126: I think the description of the ticket might be misleading. 
you don't need to change this, people can pass a role on the popover and this will be overridden.

For example, I just did that with a passing banner:
Screenshot 2022-09-21 at 9 12 41

src/popover/popover.tsx
```diff
@@ -28,7 +28,7 @@ import {filterOutFalsyProperties} from '../utils/filter-object';
 const buildContextAriaAttributes: BuildAriaAttributesFn = ({
   floating: {id, open},
 }) => ({
-  'aria-haspopup': 'dialog',
+  // 'aria-haspopup': 'dialog',
```
mutebg, line 31: You don't have to remove that, this should stay as the default behaviour.

but we need to check if people provide they aria-haspopup and if so we don't apply it automatically.

For example:
```
<Popover content="..." role="menu">
  <IconButton aria-haspopup="true" />
</Popover>
```

That logic should happen in base-floating-element.tsx file, at the point you are cloning the children you can check if the children have already aria-popup

src/popover/popover.tsx
```diff
@@ -27,8 +27,9 @@ import {filterOutFalsyProperties} from '../utils/filter-object';
 
 const buildContextAriaAttributes: BuildAriaAttributesFn = ({
   floating: {id, open},
+  ariaHasPopup,
```
mstuartf, line 30: I don't think we need to pass this down and then back up again. The purpose of `BuildAriaAttributesFn` is to expose data from the `BaseFloatingElement` that the `Popover` can't access. But the `Popover` has access to `children.props`. 

Can we simply move `buildContextAriaAttributes` inside the component and change it to:

```
const buildContextAriaAttributes: BuildAriaAttributesFn = ({
  floating: {id, open}
}) => ({
  'aria-haspopup': children.props['aria-haspopup'] || 'dialog',
  'aria-controls': open ? id : undefined,
});
```

I think this would achieve the same result without any changes to the base component. 

What do you think?

src/base-floating-element/types.ts
```diff
@@ -69,6 +80,7 @@ export interface FloatingElementProps
 
 export interface BaseFloatingElementProps extends FloatingElementProps {
   role?: AriaRole;
+  ariaHasPopup?: AriaHasPopupType;
```
mutebg, line 83: I think the idea was to check if the child already has this attribute and only if doesn't have it to be added

mutebg (approved):

mstuartf (approved):

PR #394, opened 2022-10-03
feat(PPDSC-2246): add character count component

mstuartf:
[PPDSC-2246](https://nidigitalsolutions.jira.com/browse/PPDSC-2246)

**What**

1. Background - character count component to guide users
2. What did you do:
- Add shareable refObject to form field context
- Add CharacterCount component
- Add FormInputCharacterCount component pulling required refObject from context
- Added story scenarios and tests

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


mutebg (commented):
Good one, I was thinking about the same idea of having 2 refs.

src/text-field/types.ts
```diff
@@ -36,7 +37,8 @@ type FormEntryChildrenProps = {
   onBlur: (event: React.FocusEvent<HTMLInputElement>) => void;
   onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
   state?: FormInputState;
-  ref: React.Ref<HTMLInputElement>;
+  ref: UseFormRegisterReturn['ref'];
+  refObject: React.RefObject<HTMLInputElement>;
```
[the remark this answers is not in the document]
mstuartf: Might be good to chat on a call. We get `ref` from the lib's `register` function and we have to pass it down to the input element for the functionality to work. But it's a callback, so we can't use it to access the input element elsewhere. Adding a shareable `refObject` that is also passed down allows this and shouldn't affect the behaviour of any existing `FormInput` components. 

Not sure what you mean by without a form one.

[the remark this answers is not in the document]
mstuartf: I.e. `ref` is always `UseFormRegisterReturn['ref']` because it's not created by us.

src/form/form-input.tsx
```diff
@@ -342,3 +344,10 @@ export const FormInputTextArea = React.forwardRef<
     />
   );
 });
+
+export const FormInputCharacterCount = (
+  props: Omit<CharacterCountProps, 'inputRef'>,
+) => {
+  const {refObject} = useFormFieldContext();
```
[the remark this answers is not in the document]
mstuartf: No because in a form the rules are not set on the element itself. We need to access the `rules` passed to the `FormInput`. We could require the user to duplicate the rules as `minLength` / `maxLength` attributes on the input, then we could access them using the ref but this is not ideal.

src/text-field/types.ts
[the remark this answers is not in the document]
mutebg: I am sorry, my bad, I thought refObject and ref are types for the TextField, but now I see these 2 are for FormEntryChildrenProps. 
everything is fine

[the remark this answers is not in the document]
mstuartf: 👍

src/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -46,6 +46,7 @@ Array [
   "FormController",
   "FormInput",
   "FormInputAssistiveText",
+  "FormInputCharacterCount",
```
mutebg, line 49: I think you should export the CharacterCount component as well

src/character-count/character-count.tsx
```diff
@@ -0,0 +1,97 @@
...
+      maxLengthDefault,
+    );
+    const [minLength, setMinLength] = useState<number | undefined>(
+      minLengthDefault,
+    );
+
+    const onInput: EventListener = (event: Event) => {
+      const target = event.target as ValidInputElement;
+      setCurrentLength(target.value.length);
+    };
```
mutebg, line 53: You might want to use `useCallback` for this fn

src/character-count/character-count.tsx
```diff
@@ -0,0 +1,97 @@
...
+        setCurrentLength(inputRef.current.value.length);
+        // this check ignores the browser default max length of 524,288
+        if (inputRef.current.getAttribute('maxLength')) {
+          setMaxLength(inputRef.current.maxLength);
+        }
+        // this check ignores the browser default min length of 0
+        if (inputRef.current.getAttribute('minLength')) {
+          setMinLength(inputRef.current.minLength);
+        }
+      }
```
mutebg, line 67: you need to return a function which removes the event listener

src/character-count/__tests__/character-count.stories.tsx
```diff
@@ -0,0 +1,334 @@
...
+          <FormInputLabel>Label</FormInputLabel>
+          <FormInputTextField placeholder="Placeholder" />
+          <FormInputCharacterCount />
+        </FormInput>
+      </StorybookCase>
+      <StorybookCase title="button">
+        <Button type="submit" overrides={{marginBlockStart: 'space040'}}>
+          Submit
+        </Button>
+      </StorybookCase>
```
mutebg, line 255: I guess you can combine both cases in one or set a title to an empty string, just this button label on top looks weird to me
Screenshot 2022-10-10 at 9 52 38

mutebg (commented):
That looks great!
A few small comments

src/character-count/character-count.tsx
```diff
@@ -47,25 +47,32 @@ const ThemelessCharacterCount = React.forwardRef<
       minLengthDefault,
     );
 
-    const onInput: EventListener = (event: Event) => {
+    const onInput = useCallback<EventListener>(event => {
       const target = event.target as ValidInputElement;
       setCurrentLength(target.value.length);
-    };
+    }, []);
```
mutebg, line 53: don't you need setCurrentLength as a dependency here?

src/character-count/character-count.tsx
```diff
@@ -0,0 +1,104 @@
...
+        inputEl.addEventListener('input', onInput);
+        setCurrentLength(inputRef.current.value.length);
+        // this check ignores the browser default max length of 524,288
+        if (inputEl.getAttribute('maxLength')) {
+          setMaxLength(inputRef.current.maxLength);
+        }
+        // this check ignores the browser default min length of 0
+        if (inputRef.current.getAttribute('minLength')) {
+          setMinLength(inputEl.minLength);
+        }
```
mutebg, line 68: After yesterday's conversation, I think it might be better not to use minLength/maxLength attributes from the input attributes instead, people to just pass these as props to the CharacterCount component ( same as FormCharacterCount ) in that way, we allow people to enter what-ever amount of character to the input they want without any limits.

src/character-count/__tests__/character-count.stories.tsx
```diff
@@ -0,0 +1,334 @@
...
+          <CharacterCount
+            inputRef={minHeightRef}
+            overrides={{minHeight: '80px'}}
+          />
+        </StorybookCase>
+      </StorybookPage>
+    </ThemeProvider>
+  );
+};
+CharacterCountOverrides.storyName = 'Overrides';
```
Xin00163, line 334: Do we want to add a scenario to demo the custom format? 
Scenarios' names seem to be kebab-case for other components

src/character-count/character-count.tsx
mstuartf: Do we need to remove this functionality though? Atm users can pass the min/maxLength props directly or pass the ref to the input element.

mutebg: Maybe not. probably up  to documentation to explain the differences

src/character-count/character-count.tsx
```diff
@@ -0,0 +1,107 @@
...
+      },
+      [setCurrentLength],
+    );
+
+    useEffect(() => {
+      let inputEl: ValidInputElement;
+      if (inputRef && inputRef.current) {
+        inputEl = inputRef.current;
+        inputEl.addEventListener('input', onInput);
+        setCurrentLength(inputRef.current.value.length);
```
Xin00163, line 63: We are calling `setCurrentLength` in `onInput` func then we can `setCurrentLength` again. Does those two have any differences?

src/character-count/__tests__/character-count.stories.tsx
mstuartf: What do you mean by custom format?

For the scenario names I followed the figma designs: https://www.figma.com/file/mpIRl631Oyhm585LkYzM6S/Text-Area-%26-Character-count---components?node-id=1479%3A31320. I thought perhaps we were moving to a new pattern. Can change if not though. @evgenitsn do you know what these should be?

mstuartf: P.S. I just realised what you mean by custom format @Xin00163. Good idea, will add a story.

src/character-count/character-count.tsx
mstuartf: This call is to set the initial value, then the function will be called again inside `onInput` whenever the value changes.

src/character-count/__tests__/character-count.stories.tsx
```diff
@@ -0,0 +1,334 @@
...
+  );
+};
+CharacterCountMinAndMaxLength.storyName = 'Min and max length';
+
+export const CharacterCountForm = () => (
+  <Form onSubmit={() => {}}>
+    <StorybookPage>
+      <StorybookCase title="Text area">
+        <FormInput
+          rules={{
```
Xin00163, line 213: I like how this is done rather than passing a hardcoded state. Maybe we can change other scenarios to this format later.

mstuartf: Yep, good idea.

src/character-count/__tests__/character-count.stories.tsx
```diff
@@ -0,0 +1,350 @@
...
+    nkDocs: {
+      title: 'Character count',
+      description:
+        'The Character count lets users know how much text they can enter in an input field as they type.',
+    },
+  },
+};
+
+export const CharacterCountDefault = () => {
+  const ref = useRefWithReRender<HTMLTextAreaElement>(null);
```
mutebg, line 34: Why do we need that function? Can we just use normal useRef ?

mstuartf: There's no re-render when `ref.current` is updated, so the ref passed to the CharacterCount is always undefined unless you manually trigger a state update:


> Keep in mind that useRef doesn’t notify you when its content changes. Mutating the .current property doesn’t cause a re-render. If you want to run some code when React attaches or detaches a ref to a DOM node, you may want to use a [callback ref](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) instead.
> https://reactjs.org/docs/hooks-reference.html#useref

We have a similar thing in other stories. 

Having said that, I'll look into using a callback ref to achieve this as that is what the docs recommend ☝️ . I've learned a lot about refs from this ticket!

mstuartf: Removed the null check and this function.

mutebg (approved):

PR #414, opened 2022-10-14
docs(PPDSC-2417): add theme switch to mobile menu

evgenitsn:
PPDSC-2417

**What**

1. Background - The theme toggle was missing in mobile version of the docs
2. What did you do - Added a theme toggle inside the mobile menu
3. What does the reviewers should expect - A theme switch in the mobile menu

image

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2417-add-theme-switch-to-mobile-menu
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2417-add-theme-switch-to-mobile-menu/storybook

site/components/menu-collapsible/menu-collapsible.tsx
```diff
@@ -26,6 +27,21 @@ export const GitHubLaunch: React.FC = () => (
   </StyledAnchor>
 );
 
+export const DarkModeToggle: React.FC<{
```
mutebg, line 30: isn't it a bit better if that's called theme mode toggle
```suggestion
export const ThemeModeToggle: React.FC<{
```

site/components/menu-collapsible/menu-collapsible.tsx
```diff
@@ -26,6 +27,21 @@ export const GitHubLaunch: React.FC = () => (
...
+  themeMode: string;
+  toggleTheme: () => void;
+}> = ({themeMode, toggleTheme}) => (
+  <GridLayout columns="1fr 44px">
+    <P overrides={{typographyPreset: 'utilityLabel030'}}>Dark theme</P>
```
mutebg, line 35: Is that the correct label, even when you are on the dark mode the label stays the same?
Screenshot 2022-10-17 at 10 33 46

site/components/sidebar/sidebar.tsx
```diff
@@ -1,62 +1,78 @@
...
+        <Drawer
           data-testid="sidebar"
-          role="complementary"
+          aria-label="drawer menu on the left"
+          open={sidebarOpen}
+          onDismiss={handleSidebarClick}
+          placement="left"
+          closePosition="none"
+          hideOverlay
+          inline={shouldHide}
```
mutebg, line 39: Why do you want it inline on lg and xl?

site/components/sidebar/sidebar.tsx
```diff
@@ -1,62 +1,78 @@
...
-  </>
-);
+          <Block marginInline="space060" marginBlock="space050">
+            <GitHubLaunch />
+            <Block marginBlock="space050" />
+          </Block>
+          <Block marginInline="space060" marginBlock="space050">
+            <DarkModeToggle themeMode={themeMode} toggleTheme={toggleTheme} />
+            <Block marginBlock="space050" />
+          </Block>
```
mutebg, line 59: wondering why we have these double marginBlock from parent and children

src/drawer/styled.ts
```diff
@@ -66,8 +66,5 @@ export const StyledDrawer = styled(BaseDialogView)<
         'nk-drawer',
       )(props)};
     `};
-  ${({path, inline}) =>
-    inline
-      ? logicalProps(`${path}.panel`, 'panel')
-      : logicalPaddingProps(`${path}.panel`, 'panel')};
+  ${({path}) => logicalProps(`${path}.panel`, 'panel')};
```
mutebg, line 69: this is an interesting one, I believe the idea was to not have a margin when the drawer is taking the whole screen (non-inline) but seems there is use cases for that

site/components/menu-collapsible/menu-collapsible.tsx
evgenitsn: I'll reply to this in the next comment.

site/components/menu-collapsible/menu-collapsible.tsx
evgenitsn: Yes, this is "Dark theme" by design. I think because it is a switch which can be either On or Off it makes sense to have "Dark theme" label. 
When in dark mode, it makes sense to stay "Dark theme" because the switch is **on** which means that it is currently applied. 

It wouldn't make much sense if we have "Theme mode" label or a label that will change based on the Switch state because it won't be clear what is applied and what not.

Also this one is kind of related to the comment about renaming `DarkModeToggle` to `ThemeModeToggle `. I think that since this Switch element is specifically for toggling Dark theme On and Off we are good with the `DarkModeToggle` component name. I have a less strong opinion on this one but it makes sense to me to be this way.

site/components/sidebar/sidebar.tsx
evgenitsn: Good point, I removed the inset **Block**s - `<Block marginBlock="space050" />`

src/drawer/styled.ts
evgenitsn: This is related to the "why inline on lg and xl" question above. I will try to answer that there.

site/components/sidebar/sidebar.tsx
evgenitsn: I agree, this looks confusing. It's kind of a workaround about how the `Layer` + `StyledDrawer` worked. 
The issue was that when we toggle the theme the whole Drawer disappeared after some debugging I found out that if I remove the StyledDrawer and use directly Drawer the issue is gone. This change made other two downgrades. 

- One the top margin that was applied in StyledDrawer was gone, that's why I added the `marginLogicalProps` to the Layered Drawer and using `marginBlockStart` I reverted the expected margin-top. I couldn't find any other way to add this margin because if we wrap the Drawer with another component we are getting the same issue as the initial one.
- Two the visibility hidden if the screen is LG or XL. The `Visible` component does not work by itself because the Drawer is a Layer which is a Portal and if you look at the DOM the `Visible` div element is empty, so nothing changes in terms of hiding/showing content. 
- - We have two options we can either wrap the Drawer in a component and use `getMediaQueryFromTheme('lg')` in CSS to hide it but we can't use that option because of the initial issue. The other option is to make the Drawer conditionally **inline** vs **layered** in different breakpoints. So when the screen is xs, sm and md the Drawer is inside a Layer and it works as expected and when the screen is lg or xl the Drawer is "inline" and in this way the `<Visible>` wrapper around it starts to work and basically hides it automatically. 

I agree that it is a bit confusing but it is the only way I found to make it work. Let me know if you need further explanation because I know this isn't the best one. We can chat about it. 

Screen Recording 2022-10-12 at 3 45 49 pm

mutebg: I understood, can you just add a short comment about this

mutebg: I think the other solution might be something like
<Drawer open={sidebarOpen && breakpoint !== 'lg' && breakpoint !== 'xl' />

mutebg (approved):

evgenitsn:
@mutebg @Vanals 
I've updated the Dark Theme switch to use <ScreenReaderOnly> component for the screen reader because the `label` prop is an empty string because in the design there is no label next to it. 
It seems a bit off to not use the intended label for this but the design does not use it. 
image
(Sidenote: the switch UI in the design is a bit different but that's communicated with the design team and they are ok.)

mutebg (approved):

Srinivasan-Ramasamy (approved):

Vanals (approved):

PR #425, opened 2022-10-20
docs(PPDSC-2342): added tracking to the navigation and sections clicks

evgenitsn:
PPDSC-2342

**What**

Added tracking on click for all navigations (main, side, mobile, footer) + links in main content

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


site/components/link/link.tsx
```diff
@@ -7,20 +7,26 @@ export const Link: React.FC<LinkProps> = ({
   href,
   children,
   type = 'inline',
+  eventContext = String(children),
```
mutebg, line 10: that's clever, what happens when the children are not a string but react components?

site/components/site-footer.tsx
```diff
@@ -70,42 +71,51 @@ const SiteFooter: React.FC<FooterProps> = ({cellProps = {xs: 12}}) => (
...
-          </FooterMenu>
+          <InstrumentationProvider context={{area: 'footer navigation'}}>
+            <FooterMenu>
+              <FooterLink>
+                <Link
+                  href="https://www.newscareers.co.uk/"
+                  target="_blank"
+                  external
+                  overrides={linkOverrides}
+                  eventContext={{value: 'Careers'}}
```
mutebg, line 82: Do you need to provide the eventContext here? isn't going to pick it up from the Children

mutebg (dismissed):

site/components/site-footer.tsx
evgenitsn: Correct, I left only the `Privacy policy` value because the `ConsentSettingsLink` is not picking the children.

mutebg (approved):

site/components/link/link.tsx
evgenitsn: Yeah, I wasn't covering this case. The result for these was `[object Object]` :/ . 
I added some logic around that and now the site Link is getting traversed down until it finds a string child and uses that one for the event context. I tested it on a few places, and it looks like it's working.

site/components/link/link.tsx
```diff
@@ -2,36 +2,52 @@ import React from 'react';
...
+}) => {
+  // This is traversing the children of the Link component
+  // to find the string label which is used for tracking
+  let linkLabel;
+  deepMap(children, child => {
+    if (childIsString(child)) {
+      linkLabel = child;
+    }
+    return child;
+  });
```
mutebg, line 26: do you need to do that on every render? isn't it better to be set on effect? 
Wondering if deepMap is the correct function, it can we use a filter and take the first text children?

evgenitsn: I wrapped it in a useEffect to minimise the execution. For the second part, I see what you mean, the `deepMap` doesn't fit perfectly into this logic but this is a utility that I found in [react-children-utilities.ts](https://github.com/newscorp-ghfb/newskit/blob/acdce48e720dfb1d9c6480a3b3cc75cacb3612b8/src/utils/react-children-utilities.ts). 

It's not that easy to use a filter because the Link component has nested children and we have to recursively traverse the React DOM tree to find the innermost string value which is the label. 

This `deepMap` is using `React.Children.toArray` and a few more bits into recursion to traverse the elements.

mutebg (approved):

site/pages/_app.tsx
```diff
@@ -145,6 +145,8 @@ export default class MyApp extends App<Props, State> {
           layer={{zIndex: 1000}}
           instrumentation={createEventInstrumentation(handlers, {
             ...pageProps,
+            pageOriginator: path.split('/').filter(Boolean).at(-1),
```
Xin00163, line 148: This is very nice to have.

Xin00163: The build is failing on the above line @evgenitsn 
```
Type error: Property 'at' does not exist on type 'string[]'
```

newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2324-add-tracking-to-the-navigation-and-sections-clicks
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2324-add-tracking-to-the-navigation-and-sections-clicks/storybook

site/pages/_app.tsx
evgenitsn: Thanks, should be fixed now. 🤞

Xin00163 (approved):

PR #482, opened 2022-11-21
fix(PPDSC-2554): link does not break to multiple lines

evgenitsn:
PPDSC-2554

**What**

- Dow Jones have a requirement to break links on multiple lines if they are at the end of the container width.
2. What did you do
- added `textOnly` props which when passed, we remove the `<Stack>` from the `<Link>` component and allow it to render as inline element which will wrap on multiple lines. 
- when `textOnly` is used, the end users would not be able to pass icons inside the `<Link> `children, they can achieve similar result by adding custom icons before or after the `<Link>` tag.
- updated storybook examples
- new `textOnly` prop
- updated docs 
- link using `textOnly` should wrap on multiple lines.

**Screenshot**
image

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/link/internal-link.tsx
+++ b/src/link/internal-link.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useMemo} from 'react';
 import {styled, getStylePreset} from '../utils/style';
 import {EventTrigger, useInstrumentation} from '../instrumentation';
 import {InternalLinkProps} from './types';
@@ -13,7 +13,6 @@ import {logicalProps} from '../utils/logical-properties';
 import {getTransitionPreset} from '../utils/style/transition-preset';
 
 const StyledLink = styled.a<InternalLinkProps>`
-  display: inline-block;
   ${getTransitionPreset(`link`, '')};
   ${getStylePreset('link', '')}
   ${logicalProps('link')}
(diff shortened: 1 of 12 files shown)


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2554-link-does-not-break-to-multiple-lines
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2554-link-does-not-break-to-multiple-lines/storybook

Vanals (approved):

src/link/internal-link.tsx
```diff
@@ -32,6 +32,7 @@ export const InternalLink = React.forwardRef<
     noCrop,
     href,
     external,
+    textOnly,
```
mstuartf, line 35: Could we validate that `children` is string in the `textOnly` case? Perhaps overkill.

mstuartf (approved):

mutebg (approved):

src/link/internal-link.tsx
evgenitsn: TBH, I think it would be overkill, we can't limit it easily on typescript level, we have to make checks and display warnings to the console.

mstuartf: That's fair.

PR #485, opened 2022-11-22
fix(PPDSC-2608): fix select panel position on scroll

mutebg:
PPDSC-2608

**What**

1. Background - when select panel is open and user scroll the content depending on the HTML structure ( where the scroll is ) the panel might not move together with the Selelect Button.
2. What did you do: Attachet event listener to the parent scrollable element, for that I used a helper function from `floating-ui` which find the ancestors with overflow. With that when the user scrolls, new dimensions have been taken and position updated

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


src/screen-reader-only/screen-reader-only.tsx
```diff
@@ -14,6 +14,7 @@ const ScreenReader = styled.span`
   position: absolute;
   white-space: nowrap;
   width: 1px;
+  margin: -1px;
```
mstuartf, line 17: What's the reason for this change?

mstuartf (approved):

src/screen-reader-only/screen-reader-only.tsx
mutebg: by default, the size is width: 1px; height:1px but that adds a horizontal scroll  in some cases, like imagine this screen-reader component being placed together with another element which is 100vw, this is what was happening in Dow scenario, the negative margin removes that sizing

baburay23 (approved):

PR #486, opened 2022-11-23
Chore/ppdsc 2599 fixes

mstuartf:
[PPDSC-2599](https://nidigitalsolutions.jira.com/browse/PPDSC-2599)

**What**

- Image and Select tests flaking
2. What did you do
- Enable JS on Select test that includes Tabs component, wait till images loaded before taking snapshots

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


src/image/image.tsx
```diff
@@ -117,6 +117,7 @@ const ImageComponent = React.forwardRef<HTMLImageElement, ImageProps>(
           loadingAspectRatio={loadingAspectRatio}
           ref={composeRefs(imageRef, ref)}
           src={currentSrc}
+          data-testid="loaded-image"
```
mutebg, line 120: Can we name it `image`? or can we use another selector in storybook params like `picture img` does it need to be data-testid?

mstuartf: `picture img:nth-child(1)` ensures that the loading image has disappeared:

When in loading state:
```
<picture>
   <div />
   <img />
</picture>
```

When loaded:
```
<picture>
   <img />
</picture>
```

mstuartf: Screenshot 2022-11-24 at 13 43 55
Screenshot 2022-11-24 at 13 44 00

Vanals (approved):

mutebg (approved):

PR #487, opened 2022-11-23
fix(PPDSC-2605): safari svg filters fix

evgenitsn:
PPDSC-2605

**What**

This fixes the issue with Safari where some of the SVG are not rendered correctly.
The issue is that Safari does not support additional filters like shadows, blurs and more if the svg is external i.e.imported using `<use href={...} />`.

The workaround for this is to create a new empty `<svg>` and add the `<defs>` section of the svgs (the part where the filters are located) and add this svg anywhere in the DOM.

What I did is to check if the browser is Safari, and if it is to read the svg as raw string, sanitize it, get substring of that `<defs> .. </defs>` and append it as empty, hidden svg next to the original svg. 

If the browser is not Safari, this logic is skipped and nothing changes. 
Looks like this approach fixes the Safari issues with the illustrations.

Example of the DOM for Safari vs Chrome

Safari:
image

Chrome:
image

**I have done:**

**I have tested manually:**

**Before:**

image

**After:**
image

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2605-svg-shadow-elements-missing-in-safari
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2605-svg-shadow-elements-missing-in-safari/storybook

package.json
```diff
@@ -188,6 +188,7 @@
     "package-change-checker": "^1.2.4",
     "prettier": "2.2.1",
     "pretty-proptypes": "^0.6.0",
+    "raw-loader": "^4.0.2",
```
mutebg, line 191: Do we need to add this to ignore dependencies list?

site/components/illustrations/illustration-loader.tsx
```diff
@@ -1,5 +1,34 @@
...
-import {pathToID} from './utils';
+import dompurify from 'isomorphic-dompurify';
+import {isSafari, pathToID} from './utils';
+
+/**
+ * Only for Safari. It reads the svg file as raw string and returns the problematic <defs> .. </defs> part as additional svg which solves the problem. If the browser is not Safari then all of this is skipped.
+ */
+const loadSVGFiltersForSafari = (path: string) => {
+  if (!isSafari) return null;
+  const {sanitize} = dompurify;
```
mutebg, line 10: wondering if we can load the dompurify with dynamic import so that our bundle size does not increase

package.json
evgenitsn: This is a `devDependency` I believe it will be omitted automatically. The `dependenciesNotToBeIncluded` list does contains only `dependencies` and not `devDependency`. Let me know if I am missing something.

site/components/illustrations/illustration-loader.tsx
evgenitsn: In this case we can, but we are already using the dompurify in another Component: `site/components/tools/svg-previewer.tsx`. So I think the bundle size is already including that. 

I tried to import it dynamically there as well but I wasn't able to figure out how to dynamically import an npm package in a react component body. If you have any ideas, happy to discuss.

mutebg: there is a difference I believe, there is a bundle per page, so the previewer will include it only there, and since we start using it for SVG's and they are on all pages it will start to be included everywhere.
I think you might need to do that inside useEffect and maybe create a custom hook in order to achieve it. I am not sure if its with the time these extra 10kb but have a try

package.json
mutebg: You are right, I did not see it

site/components/illustrations/illustration-loader.tsx
evgenitsn: Updated the code, the sanitizer should be load dynamically only for Safari now.

site/components/illustrations/illustration-loader.tsx
```diff
@@ -1,18 +1,78 @@
...
+const getSafariSVGFilters = (
+  sanitize: (source: string) => string,
+  path: string,
+) => {
+  const rawSVG = sanitize(
+    // eslint-disable-next-line import/no-dynamic-require, global-require
+    require(`../../public/static/illustrations/${path}.svg`).default,
+  );
+
+  const defsStart = '<defs>';
```
mstuartf, line 16: Could there be any attributes in this opening tag that might mean the match doesn't work?

site/components/illustrations/illustration-loader.tsx
```diff
@@ -1,18 +1,78 @@
...
   const id = pathToID(path);
 
-  const Component = () => (
+  const renderInlineSVG = () => (
     <svg viewBox="0 0 1490 838" width="100%" {...props}>
       <use href={`static/illustrations/${path}.svg#${id}`} />
     </svg>
   );
 
+  const Component = () => {
```
mstuartf, line 44: Can we add tests for this logic?

site/components/illustrations/illustration-loader.tsx
LukeFinch: <defs> can have "lang" and "id" they're also technically styleable.
Might be worth doing a regex to match defs just in case the export from figma adds anything in there.

`<defs[^(\/>)]*>` this should catch any opening def tag

mstuartf:
Screenshot 2022-11-25 at 10 39 18
Screenshot 2022-11-25 at 10 39 22
@evgenitsn fyi it looks like this is being picked up by the Percy tests, so you will have some to resolve.

evgenitsn:
> Screenshot 2022-11-25 at 10 39 18
> 
> Screenshot 2022-11-25 at 10 39 22
> 
> @evgenitsn fyi it looks like this is being picked up by the Percy tests, so you will have some to resolve.

Thanks Mike, yes I noticed this in a few more cases. I will push an update. Marking it as WIP now as the implementation would change a bit.

evgenitsn:
I've updated the implementation, it turned out that appending only the <defs> </defs> section is not reliable solution because not all filters are contained inside it. 

So currently for Safari we are just inlining the SVG into the code (same as before the svg/tsx change) and for the rest of the browsers we are using external SVGs inside the <use/> i.e. how it is supposed to work.

site/components/illustrations/illustration-loader.tsx
evgenitsn: I tried to add testing to this, but I am facing issues with the jest SVG tranform and also I get weird errors which I guess are because of the dynamic import because it's an async operation so I am a bit stuck at the moment.

If anyone keen to assist we can try to solve these. I have a feeling that this task already took a lot of time so idk :/

evgenitsn:
Also there is one failing e2e test which on a first looks shouldn't be involved with my changes. I am having issues reproducing it locally.

site/components/illustrations/illustration-loader.tsx
evgenitsn: Thanks @LukeFinch I left a comment below that this `<defs..>` section is not enough for all illustrations so I've updated the code to inject inline the whole svg only for Safari. So this comment is not relevant anymore.

site/components/illustrations/illustration-loader.tsx
mstuartf: OK, maybe we can add in the future

evgenitsn:
@mstuartf @mutebg @LukeFinch Thank for looking into this. I think it's in a good shape now, ready for reviews.

site/components/illustrations/utils.ts
```diff
@@ -33,3 +33,11 @@ export const pathToID = (path: string) => {
...
   return '';
 };
+
+const safariUACheck = /^((?!chrome|android).)*safari/i;
+const isWindowLoaded = typeof window !== 'undefined';
+
+export const isSafari =
+  isWindowLoaded && safariUACheck.test(window.navigator.userAgent);
+export const isNotSafari =
+  isWindowLoaded && !safariUACheck.test(window.navigator.userAgent);
```
LukeFinch, line 43: What's the reason for having both? does `!isSafari` not satisfy the need?

LukeFinch: Fair enough, I figured it would have to do with the window being loaded. 👍

LukeFinch (dismissed):

site/components/illustrations/illustration-loader.tsx
```diff
@@ -1,17 +1,69 @@
...
+        viewBox="0 0 1490 838"
+        width="100%"
+        opacity={isNotSafari ? 1 : 0}
+        {...props}
+      >
+        <use href={`static/illustrations/${path}.svg#${id}`} />
+      </svg>
+    );
+
+    return safariSVG === null ? nonSafariSVG : renderSafariSVG;
```
mstuartf, line 65: Can this be simplified?

I think there is duplicate logic:
* `safariSVG` will be null unless `isSafari` is true due to the if condition in the `useEffect`
* But you are also checking `isSafari` again when defining `renderSafariSVG`

Could you just remove `renderSafariSVG` and make the final return statement:
`return safariSVG || nonSafariSVG;`

Or do you want `nonSafariSVG` to render first, then to be replaced with `safariSVG` when the async logic finishes?

site/components/illustrations/illustration-loader.tsx
```diff
@@ -1,17 +1,69 @@
...
+        // Dynamically importing 'isomorphic-dompurify' only for Safari to avoid bloating the bundle size for other browsers.
+        const importSanitizer = async () => {
+          const dompurify = await (await import('isomorphic-dompurify'))
+            .default;
+          const svg = getSafariSVG(dompurify.sanitize, path, props);
+          setSafariSVG(svg);
+        };
+        importSanitizer();
+      }
+    }, [safariSVG]);
```
mstuartf, line 48: Could this just run on first render with `[]` as dependencies, and no need to check `safariVG === null`?

site/components/illustrations/illustration-loader.tsx
evgenitsn: Just tested this one, it works fine and it looks much cleaner, thanks. Some of these checks were leftover from the previous approach using `<defs>` substring because we had to have more control over the order of rendering but with the current approach we can simplify. 👍

site/components/illustrations/illustration-loader.tsx
evgenitsn: Done

mstuartf (dismissed):

LukeFinch (approved):

LukeFinch (dismissed):

site/components/tools/svg-previewer.tsx
```diff
@@ -1,6 +1,6 @@
 import React, {useEffect, useState} from 'react';
 import {getColorCssFromTheme, getSSRId, P} from 'newskit';
-import dompurify from 'dompurify';
+import dompurify from 'isomorphic-dompurify';
```
Vanals, line 3: Hey! why is this needed 🤔 ?

evgenitsn: If you are asking if the library change from `dompurify` to `isomorphic-dompurify` is needed. I added that because I need `isomophic-dompurify` into another component and decided to swap it in both places to not keep two libs for almost the same thing.

If you are asking if we need dompurify lib for this `svg-previewer.tsx` file, I am not sure, I haven't spend much time looking into this file and what it is doing, just replaced the lib. So can't really tell.

LukeFinch (approved):

Vanals (approved):

PR #502, opened 2022-11-30
docs(PPDSC-2595): Use Google Sheets as CMS for Roadmap page

LukeFinch:
PPDSC-2595

**What**

Used for local data: https://docs.google.com/spreadsheets/d/12MSFqNHH3pql72fpGQh5GRNFUn22SFYQKtgIo0I0nCQ/edit#gid=1318720245

- We would like to be able to manage the roadmap page by providing the data to the page through a google sheet
2. What did you do
- Set up dotenv so we can develop with google sheets locally. (See LastPass for credentials)
- Proposed a document structure for the google sheet

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2595-roadmap-page
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2595-roadmap-page/storybook

package.json
```diff
@@ -223,6 +223,7 @@
     "debounce": "^1.2.0",
     "dequal": "^1.0.0",
     "dompurify": "^2.3.1",
+    "dotenv": "^16.0.3",
```
mutebg, line 226: Any new dependency needs to be added to `scripts/build-package-json.js` dependenciesNotToBeIncluded list so that   is not included in newskit package

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,196 @@
...
+    </AboutPageTemplate>
+  );
+};
 
 export default Roadmap;
+
+// This function is called at build time and the response is passed to the page
+// component as props.
+export async function getStaticProps() {
+  const [cmsData] = await Promise.all([getSheets('Roadmap')]);
```
mutebg, line 192: In this case, we have only 1 async task so no promise.all is needed
```suggestion
  const cmsData = await getSheets('Roadmap');
```

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,196 @@
...
+            {children}
+          </LinkStandalone>
+        ),
+        p: ({children}) => <>{children}</>,
+      }}
+    >
+      {children}
+    </ReactMarkdown>
+    /* eslint-enable @typescript-eslint/no-shadow */
+  );
```
mutebg, line 51: Can we move this component out of that scope so no need to be recreated on every render?

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,196 @@
...
+            <UnorderedList
+              overrides={{
+                content: {
+                  typographyPreset: 'editorialParagraph020',
+                  stylePreset: 'inkBase',
+                },
+              }}
+            >
+              {cmsData
+                .filter(data => data[0].startsWith('current_li'))
```
mutebg, line 115: IMO  in the Sheet, we should just use only `current_li` instead of `current_li_X`

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,196 @@
...
+                  typographyPreset: 'editorialParagraph020',
+                  stylePreset: 'inkBase',
+                },
+              }}
+            >
+              {cmsData
+                .filter(data => data[0].startsWith('comingup_li'))
+                .map(entry => (
+                  <FormatMarkdown>{entry[1]}</FormatMarkdown>
+                ))}
```
mutebg, line 150: Maybe can help a helper function for that?
```
const getCMSList = (cmsData, listKey) => cmsData
        .filter(data => data[0].startsWith(listKey))
        .map(entry => (
          <FormatMarkdown>{entry[1]}</FormatMarkdown>
        ))}
```
and `getCMSList(cmsData, 'comingup_li');`

mutebg (commented):
Good job, few small comments

package.json
LukeFinch: Thanks for the heads up. I'll make sure its in there.

site/pages/about/roadmap.tsx
LukeFinch: Ideally, every key in the sheet should be unique, so I've numbered them. Are you thinking it could be `current_liX` not `current_li_X` ?

site/pages/about/roadmap.tsx
LukeFinch: This is a good idea 👍

site/pages/about/roadmap.tsx
mutebg: I meant, do we need to number them at all?

LukeFinch: https://github.com/newscorp-ghfb/newskit/blob/a4c37e1d860536ab701cad77fabaa3c9b270b119/site/utils/google-sheet/utils.ts#L15-L26

the formatSheetData function wouldn't necessarily parse them correctly- duplicate keys will overwrite each other.
We could allow keys to be the same, and then add them to an array, by changing the function a bit

It would save me from having to search the data directly for those keys and filtering.

mutebg: IMO it will make it easier for people to enter data in a sheet and not have to worry about numbering every single item, but at the same time, I don't know if is worth the effort. It's up to you I guess :) We always can change it if people start making mistakes with these numbers

LukeFinch: Yeah, I get what you mean...
Part of me thinks it would be better to keep all the keys unique so that we can go with the rule "all keys in the sheet need to be unique". 
Sheets will automatically number it if. you drag down the row above - shouldn't be too bad hopefully.

mutebg (approved):

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,191 @@
...
+  );
+};
+
 export default Roadmap;
+
+// This function is called at build time and the response is passed to the page
+// component as props.
+export async function getStaticProps() {
+  const cmsData = await getSheets('Roadmap');
+  const content = formatSheetData(cmsData);
```
mstuartf, line 189: It's a bit confusing to pass the same data in two different structures to the component. 

If we want to stick with this pattern perhaps renaming the variables might be a bit clearer. E.g. `cmsDataArr` and `cmsDataObj`. I would also do this transformation inside the component and just do the API call inside of `getStaticProps`. 

What do you think? Is there one preferred structure we could use or do we need both?

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,191 @@
...
+    'intro_description',
+    'NewsKit’s Design System team is busy building and planning to help you build better products faster.',
+  );
+
+  const introSecondary =
+    cmsData.find(data => data[0] === 'intro_secondary') &&
+    getValueFromCMS(
+      content,
+      'intro_secondary',
+      'The roadmap is a living document, and it is likely that priorities will change. See our Trello board for more details on the roadmap.',
```
mstuartf, line 58: Out of scope, but what do we think about having a default object in the page dir with all of the fallback strings? Sort of similar to the default styling objects we have for components. We could merge this object with data fetched from the API at the start, rather than passing defaults to every `getValueFromCMS` call.

site/pages/about/roadmap.tsx
LukeFinch: Agree it feels a little bit gross.

I could keep the Object, and then use `Object.entries()` when I need to do array methods on it. 

The preferred structure - if we look at the homepage example is to go with the object, because the general pattern is key-value-pair. In this instance, I'm using the array to loop through some of the keys to generate a list.

mstuartf: I think using the object makes sense - and perhaps the transformation could be part of `getSheets`. Then we could just update the `getCMSList` function to expect an object and do the `Object.entries` call internally.

site/pages/about/roadmap.tsx
mstuartf: 👍 Do you want to do as part of this ticket or later on?

site/utils/google-sheet/utils.ts
```diff
@@ -1,15 +1,9 @@
 import {CMSDataProps} from './types';
 
-export const getValueFromCMS = (
-  obj: CMSDataProps,
-  key: string,
-  fallback: string,
-) => {
+export const getValueFromCMS = (obj: CMSDataProps, key: string) => {
   const newKey = key.toLowerCase();
```
mstuartf, line 4: Would it be better to just type the CMS object for each page?

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,193 @@
...
-            details.
+const getCMSList = (content: CMSDataProps, listKey: string) =>
+  Object.entries(content)
+    .filter(entry => entry[0].startsWith(listKey))
+    .map(entry => <FormatMarkdown>{entry[1]}</FormatMarkdown>);
+
+const Roadmap = ({
+  content,
+  ...layoutProps
+}: LayoutProps & ContentProps & CMSDataProps) => {
```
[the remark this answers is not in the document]
mstuartf: Or just `LayoutProps & {content: RoadmapContent}`.

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,193 @@
...
+  const pageName = getValueFromCMS(content, 'intro_name');
+  const pageDescription = getValueFromCMS(content, 'intro_description');
+
+  const introSecondary =
+    content.intro_secondary && getValueFromCMS(content, 'intro_secondary');
+
+  if (Object.entries(content).length) {
```
mstuartf, line 56: This check will always pass due to the defaults. Should it be checking the dynamic keys?

site/pages/about/roadmap.tsx
```diff
@@ -33,31 +68,120 @@ const FormatMarkdown: React.FC<ReactMarkdownOptions> = ({children}) => (
   /* eslint-enable @typescript-eslint/no-shadow */
 );
 
-const getCMSList = (cmsData: string[][], listKey: string) =>
-  cmsData
-    .filter(data => data[0].startsWith(listKey))
+// TODO: Types for keyof starting with...
```
mutebg, line 71: Do you need this comment?

site/pages/about/roadmap.tsx
```diff
@@ -33,31 +68,120 @@ const FormatMarkdown: React.FC<ReactMarkdownOptions> = ({children}) => (
...
   ...layoutProps
-}: LayoutProps & ContentProps & {cmsData: string[][]}) => {
-  const pageName = getValueFromCMS(content, 'intro_name', 'Roadmap');
-  const pageDescription = getValueFromCMS(
-    content,
-    'intro_description',
-    'NewsKit’s Design System team is busy building and planning to help you build better products faster.',
-  );
+}: LayoutProps & {content: RoadmapContent}) => {
+  const pageName = content.intro_name; // getValueFromCMS(content, 'intro_name');
```
mutebg, line 82: this one as well

mutebg (approved):

mutebg (approved):

mutebg (dismissed):

site/pages/about/roadmap.tsx
```diff
@@ -1,68 +1,246 @@
...
-              border: '1px solid #ccc',
-            }}
-          />
-        </ContentPrimary>
+    {children}
+  </ReactMarkdown>
+  /* eslint-enable @typescript-eslint/no-shadow */
+);
+
+const getCMSList = (content: RoadmapContent, listKey: keyof RoadmapContent) =>
```
mstuartf, line 79: I would make this a generic function with a type arg so you can use it on other pages.

E.g.  `const getCMSList = <T>(content: T, listKey: keyof T) =>`

mstuartf (approved):

mstuartf (approved):

site/utils/google-sheet/get-sheets.ts
```diff
@@ -24,7 +26,8 @@ export async function getSheets(range: string) {
     });
     return response.data.values;
   } catch (err) {
-    console.log('>> ERROR: Cannot fetch data from googlesheet api');
+    console.error('>> ERROR: Cannot fetch data from googlesheet api');
+    console.log('Have you added the .env file for local builds?');
```
Vanals, line 30: should this be a console.warn maybe!?  or part of the error itself. Maye more easly get lost in the logs I think. Also, should we not consol error the `err` it self? so we know what the issue is. Let say depends from them.. we migth receive a clue

Vanals (approved):

PR #525, opened 2022-12-19
fix(PPDSC-2596): flexbox issue in Safari

Xin00163:
PPDSC-2596
- Change `end` & `start` to `flex-end` & `flex-start`

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2596-flex-safari-bug
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2596-flex-safari-bug/storybook

JohnTParsons (dismissed):

src/test/storybook-comps.tsx
```diff
@@ -143,7 +143,7 @@ export const StorybookCase = ({
 );
 
 export const StoryDocsHeader = ({context}: {context: DocsContextProps}) => {
-  const autoTitle = get(context, 'title').split('/').at(-1);
+  const autoTitle = context.title && context.title.split('/')[-1];
```
evgenitsn, line 146: I think this `[-1]` introduces a bug, if you have a look at the not updated components for example [fieldset on prod](https://storybook.newskit.co.uk/?path=/docs/components-fieldset--story-fieldset-default) you can see that `fieldset` is extracted as `autoTitle` and used for title on the page, whereas in [this PR the fieldset is empty](http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2596-flex-safari-bug/storybook/?path=/docs/components-fieldset--story-fieldset-default) because `[-1]` returns `undefined` because it looks for element on index `-1` vs `.at(-1)` which looks at the last array element.
Reference: (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#examples)

Xin00163: hmm... that's why we use `at`. Thanks @evgenitsn. Yeah I remembered this is due to my safari being old now. With the latest version, should be OK

evgenitsn: Yeah, should be good for Safari 15.4 and newer, the other good old approach is to use `[arr.length-1]` but I think `.at()` has pretty good support.
https://caniuse.com/?search=Array.prototype.at

mutebg (approved):

PR #544, opened 2023-01-05
docs(PPDSC-2653): fix console errors on homepage

evgenitsn:
PPDSC-2653

**What**

1. Background - There was a few console errors and warning on the docs home page 
2. What did you do - Fixed them
3. What does the reviewers should expect - No errors in the browser console on home page

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


newskit-engineering:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2653-fix-console-errors-on-homepage
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2653-fix-console-errors-on-homepage/storybook

JohnTParsons (dismissed):

src/audio-player-composable/components/volume-control/volume-control.tsx
```diff
@@ -182,6 +183,7 @@ const ThemelessAudioPlayerVolumeControl = React.forwardRef<
       {useSliderContainer && (
         <GridLayoutItem area="slider">
           <CSSTransition
+            nodeRef={cssTransitionNodeRef}
```
mutebg, line 186: can you explain what's going on here, why do we need that on every component that has CSS transition?

src/audio-player-composable/components/volume-control/volume-control.tsx
```diff
@@ -193,15 +195,17 @@ const ThemelessAudioPlayerVolumeControl = React.forwardRef<
...
-              <StyledVolumeSliderContainer
-                className={getTransitionClassName('nk-vc', state)}
-                layout={layout}
-                overrides={overrides}
-                open={open}
-                visible={isVisible}
-              >
-                {sliderComponent}
-              </StyledVolumeSliderContainer>
+              <div ref={cssTransitionNodeRef}>
```
mutebg, line 198: can you pass the ref directly to StyledVolumeSliderContainer? I personally don't like adding extra div elements

mutebg (changes requested):

src/audio-player-composable/components/volume-control/volume-control.tsx
evgenitsn: There was an error in the console 

> Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node

After some research I found out that it is coming from react-transition-group package and this is the issue and the proposed solution. 
https://github.com/reactjs/react-transition-group/issues/668#issuecomment-695162879

This is also present in their official docs (first example): https://reactcommunity.org/react-transition-group/css-transition

src/audio-player-composable/components/volume-control/volume-control.tsx
evgenitsn: Following the my other comment, we have to have the `ref` prop to avoid the error in the console. It may be a good idea to try and pass it to the outermost node in this case `StyledVolumeSliderContainer`. I will check that and will update where is possible.

mutebg: Can you do that for all components where you pass an extra div, I don't see why wound't work

mstuartf: Solution looks good but I agree we should just be able to pass to the existing element, rather then wrapping in a new div.

evgenitsn: Thanks for the review, I've just updated the components. No more wrapping divs, just passing the ref to the original parent. @mutebg @mstuartf

mutebg (approved):

baburay23 (approved):

PR #583, opened 2023-01-24
fix(PPDSC-2717): fix aria-labelledby on select panel

mutebg:
PPDSC-2717

**What**

1. Background - bug in our component
2. What did you do - 
   - added `labelId` prop, since the select doesn't know anything about the label we need to provide the id of the label to the select-panel. In cases labelId is not provided aria-labellledby will not be rendered.
   - reversed some of the changes in PPDSC-2661 related to React. children.toArray which breaks downshift.
   - updated select itemToString so that ScreenReader reads the SelectOption text when there is no aria-label  ( before that was reading the value which does not seems correct )

Before:
Screenshot 2023-01-24 at 12 56 03

After:
Screenshot 2023-01-24 at 12 55 24

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


pp-serviceaccount:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2717-select-aria-labelledby-issue
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2717-select-aria-labelledby-issue/storybook

site/components/svg-previewer/__tests__/__snapshots__/svg-previewer.test.tsx.snap
```diff
@@ -2409,7 +2409,6 @@ Object {
         class="nk-layer"
       >
         <div
-          aria-labelledby="downshift-0-label"
```
Vanals, line 2412: Hey  I was having a look trying to get an idea. We have removed the aria-labelledby because it is not needed right?

But we have added a prop so they can add it back and be linked properly. If correct, which benefits is bringing adding the aria labelled by? in terms of ScreenReader.  Happy to chat if did not ge tit right

src/select/__tests__/select.stories.tsx
```diff
@@ -981,11 +981,14 @@ export const StorySelectControlled = () => {
   return (
     <Container>
       <Block>
-        <Label htmlFor="controlled">Controlled</Label>
+        <Label htmlFor="controlled" id="controlled-label">
```
Vanals, line 984: Not sure if was like that before as well, I tried to remove the props and seems to be, but if feels a bit broken when scanning it with the SR on safari

src/form/form-input.tsx
```diff
@@ -197,6 +197,7 @@ export const FormInputSelect = React.forwardRef<HTMLInputElement, SelectProps>(
         onBlur={composeEventHandlers([onBlur, onBlurContext])}
         ref={composeRefs(ref, inputRef)}
         id={id}
+        labelId={labelId}
```
Xin00163, line 200: I wondered if we need labelId here? It seems to always come from the context.

mutebg: I am not sure I get your question.
`labelId` is passed from the context to Select component

Xin00163 (approved):

Vanals (dismissed):

Vanals (approved):

Xin00163 (approved):

mstuartf (approved):

PR #653, opened 2023-02-20
feat(PPDSC-2798): expose css variables

mutebg:
PPDSC-2798

**What**

2. What did you do - added CSS variables to theme provider

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


pp-serviceaccount:
  Documentation Site: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2798-expose-css-variables
  Storybook: http://ncu-newskit-docs.s3-website-eu-west-1.amazonaws.com/ppdsc-2798-expose-css-variables/storybook

LukeFinch:
All looks good, just a few questions:

- Do we want to namespace our variables to avoid potential clashes with other libraries?
- Have we considered data-attributes rather than class names?
  - something like data-theme="newskitLightTheme"

mutebg:
> All looks good, just a few questions:
> 
> * Do we want to namespace our variables to avoid potential clashes with other libraries?
That's a good one, however how likely is The Sun to use other UI libraries at the same time?

> * Have we considered data-attributes rather than class names?
>   
>   * something like data-theme="newskitLightTheme"
Can you elaborate on that, I don't understand it

LukeFinch (dismissed):

site/pages/_app.tsx
```diff
@@ -156,6 +155,7 @@ export default class MyApp extends App<Props, State> {
 
         <NewsKitProvider
           theme={theme}
+          themeOptions={{exposeCSSVariables: true}}
```
JohnTParsons, line 158: Would `exposeCSSVariables` be better camelCase if called `exposeCssVariables`?

mutebg: I don't know really, from one side camelcase is more readable, on the other CSS is an acronym and as such, it needs to be upper-case.  Happy to change it if you think that's better

Xin00163 (commented):
Nice work. Tested a few scenarios and all look good

src/theme/__tests__/theme-provider.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+        1st level
+        <ThemeProvider theme={childrenTheme} exposeCSSVariables>
+          2nd level
+        </ThemeProvider>
+      </ThemeProvider>,
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it.only('themeDiff -> different token', () => {
```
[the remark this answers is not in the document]
mutebg: Thanks, fixed

Xin00163 (approved):

site/pages/_app.tsx
JohnTParsons: I'm inclined to think that exposeCssVariables is better

JohnTParsons (approved):

Xin00163 (approved):

mstuartf:
I think @LukeFinch's point about namespacing is a good one. But I think it's fine for now. If it turns out to be something we need, we could add support for an optional namespace prefix prop to be passed in along with `exposeCssVariables`.

mstuartf (approved):

mutebg:
> I think @LukeFinch's point about namespacing is a good one. But I think it's fine for now. If it turns out to be something we need, we could add support for an optional namespace prefix prop to be passed in along with `exposeCssVariables`.

Thanks, I totally agree, let's see if anyone needs that we can easily add it as an additional option to the theme provider.

Xin00163 (dismissed):

LukeFinch (dismissed):
Nice work 👏

JohnTParsons (dismissed):

Issue #690, opened 2023-03-07
Option to have `full-width` cell in structuredListItem when there is only one child


Xin00163:
# Is your feature request related to a problem? Please describe.
When there is only one child, the user would like to remove the padding around the cell. Currently this is not doable in structuredListCell because it's using `xs={12}`. 

# Describe the solution you'd like
Ideally we should have an option to pass `xs='full-width'`so the padding will be removed.

# Describe alternatives you've considered
*A clear and concise description of any alternative solutions or features you've considered.*
Add a prop like below
```
      if (childrenArray.length === 1) {
        return (
          <StyledGrid
            xsRowGutter="space000"
            xsMargin="space020"
            lgMargin="space030"
            overrides={overrides}
            hasHref={hasHref}
          >
            <StyledCell xs={fullWidthCell? 'full-width': 12} align={alignCell1}>
              {childrenArray[0]}
            </StyledCell>
          </StyledGrid>
        );
      }
```

# Additional context
Copy the below scenario in storybook to view the issue
```
        <StructuredList
          ariaLabel="list"
          overrides={{
            paddingInline: 'space000',
            paddingBlock: 'space000',
            width: 'unset',
          }}
        >
          <StructuredListItem
            ariaLabel="list item"
            overrides={{
              paddingInline: 'space000',
              paddingBlock: 'space000',
            }}
          >
            <StructuredListCell>
              <HeadingTextBlock>[Cell 1]</HeadingTextBlock>
            </StructuredListCell>
          </StructuredListItem>
        </StructuredList>
```

PR #693, opened 2023-03-08
feat(ISSUE-4): add fullWidthSingleCell prop to structuredListCell

Xin00163:
https://github.com/newscorp-ghfb/newskit/issues/690

**What**

1. Background - As a user, I would like to align the divider and content when there is only one cell in StructuredList so there is no extra padding/margin
2. What did you do - to avoid making a breaking change, I've added a prop to allow this to happen
3. What does the reviewers should expect - The rest of the scenarios should still work as expected. There is one extra scenario added to one cell.

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/structured-list/types.ts
+++ b/src/structured-list/types.ts
@@ -44,4 +44,5 @@ export interface StructuredListCellProps extends CellProps {
   children: React.ReactNode;
   pullRight?: boolean;
   align?: StructuredListCellAlign;
+  fullWidthSingleCell?: boolean;
 }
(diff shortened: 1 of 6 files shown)


src/structured-list/types.ts
```diff
@@ -44,4 +44,5 @@ export interface StructuredListCellProps extends CellProps {
   children: React.ReactNode;
   pullRight?: boolean;
   align?: StructuredListCellAlign;
+  fullWidthSingleCell?: boolean;
```
pdimova, line 47: 👍

mutebg (approved):

Vanals (approved):

Issue #709, opened 2023-03-15
(Bug) - `StructuredList` with `Divider` has an a11y issue


evgenitsn:
# Describe the bug
The `StructuredList` component when rendered with `divider` prop has an accessibility issue.
It renders `<hr/>` elements as direct children of `<ul>` next to `<li>`s which is not correct.  

https://dequeuniversity.com/rules/axe/4.4/list

This gets triggered by the axe plugin only if we pass `role="list"` prop to the `<StructuredList>` and that's why is doesn't show as  an issue on the newskit storybook. If you pass a `role` to any example with dividers it should start complaining.

# To Reproduce
1. Render a `StructuredList` with divider and pass a `role="list"` to it.
2. Open and scan the whole page with `axe DevTools` extension
3. See "Certain ARIA roles must contain particular children" error

> To solve this problem, you need to fix at least (1) of the following:
Element has children which are not allowed (see related nodes)
Element has no aria-busy="true" attribute

Here is a code snippet of how we use the `StructuredList`
```tsx
          <StructuredList
            divider={showDivider}
            ariaLabel={ariaLabel || introductionProps?.title}
            {...{ role: 'list' }}
          >
            {children as any}
          </StructuredList>
```
Link to the source: https://github.com/newscorp-ghfb/ncu-newskit-render/blob/master/packages/shared-components/src/ContentListView/ContentListView.tsx#L63-L69

# Expected behaviour
An a11y error should not show. The `ul` can't have direct `<hr>` children. 
A proposal to fix this could be to wrap the `<hr>` elements in `<li>` elements, hide them and move the `aria-hidden` from the `hr` to the wrapping `li` elements.  

# Screenshots / Video / Test Environment
<img width="868" alt="image" src="https://user-images.githubusercontent.com/7677924/225251893-d7d1d44c-3175-4f09-b7ff-15b00c6855c1.png">

# Desktop (please complete the following information):
 - OS: macOS Ventura
 - Browser Chrome
 - Version 110.0.5481.177

# Additional context
*Add any other context about the problem here.*

PR #710, opened 2023-03-15
fix(SCS-70): structured-list with divider a11y issue

evgenitsn:
[SCS-70](https://nidigitalsolutions.jira.com/browse/SCS-70)

**What**

1. Background - #709 
2. What did you do - wrapped the Divider elements in list items 
3. What does the reviewers should expect - no a11y errors even if the list has a `role="list"`

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

[SCS-70]: https://nidigitalsolutions.jira.com/browse/SCS-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/structured-list/__tests__/__snapshots__/structured-list.test.tsx.snap
+++ b/src/structured-list/__tests__/__snapshots__/structured-list.test.tsx.snap
@@ -44,11 +44,15 @@ exports[`StructuredList with href renders four cells with href 1`] = `
         href="http://apple.com"
       />
     </li>
-    <hr
+    <li
       aria-hidden="true"
-      class="emotion-3"
-      data-testid="divider"
-    />
+    >
+      <hr
+        aria-hidden="true"
+        class="emotion-3"
+        data-testid="divider"
+      />
+    </li>
     <li
       aria-label="list item"
       class="emotion-1"
@@ -266,11 +270,15 @@ exports[`StructuredList with href renders one cell with disabled link and hidden
         </div>
       </span>
     </li>
-    <hr
+    <li
       aria-hidden="true"
-      class="emotion-8"
-      data-testid="divider"
-    />
+    >
+      <hr
+        aria-hidden="true"
+        class="emotion-8"
+        data-testid="divider"
+      />
+    </li>
     <li
       aria-label="list item"
       class="emotion-1"
@@ -509,11 +517,15 @@ exports[`StructuredList with href renders one cell with external link and hidden
         </div>
       </a>
     </li>
-    <hr
+    <li
       aria-hidden="true"
-      class="emotion-8"
-      data-testid="divider"
-    />
+    >
+      <hr
+        aria-hidden="true"
+        class="emotion-8"
+        data-testid="divider"
+      />
+    </li>
     <li
       aria-label="list item"
       class="emotion-1"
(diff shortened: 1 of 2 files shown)


JohnTParsons (approved):

pp-serviceaccount:
  Documentation Site: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/scs-70-structured-list-dividers-a11y-issue
  Storybook: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/scs-70-structured-list-dividers-a11y-issue/storybook

src/structured-list/__tests__/__snapshots__/structured-list.test.tsx.snap
```diff
@@ -44,11 +44,15 @@ exports[`StructuredList with href renders four cells with href 1`] = `
...
     </li>
-    <hr
+    <li
       aria-hidden="true"
-      class="emotion-3"
-      data-testid="divider"
-    />
+    >
+      <hr
+        aria-hidden="true"
```
mutebg, line 51: if the parent has aria-hidden, the children hr, in that case, does not need it. 
but looking into Divider I don't think is something we can support.
Have you checked if that's ok with axe and SR?

evgenitsn: Yes, we can't remove the `aria-hidden` from Divider in its current implementation.
I've added `role="list"` to an example for test and there were no errors on Axe and the screen reader is navigation smoothly through the list, no mentioning for dividers or any other unwanted elements.

Vanals:
Shall we add also role='list' in storybook scenarios? so we can capture this or other faulty scenarios.

src/structured-list/__tests__/__snapshots__/structured-list.test.tsx.snap
evgenitsn: I think it would be hard to do it like this
```
 <li> 
     Item
     Divider
  </li>
```
because we get the items like this 
```
const structuredListChildren = React.Children.toArray(
    children,
  ) as React.ReactElement<StructuredListItemProps>[];
```
so the children are in the hands of the users and we can't easily determine where to put the divider.

I get your point but it doesn't look like a straightforward thing to do and for the less DOM nodes point, isn't is the same, the Divider is still an `hr` element, the amount of nodes should be the same, the `hr` will just be nested?

evgenitsn:
> Shall we add also role='list' in storybook scenarios? so we can capture this or other faulty scenarios.

Don't know to be honest, the whole component has the role - list implicitly either way. So in theory no one should use `role="list"` in their usage. Also the `StructuredList` does not accept `role` prop according to the types. I think on `render` where I found this usage they were forcing it (ignoring the types). 

image

Not sure why this a11y error is showing only after adding the role. If we want an example of that we would have to extend the StructuredList props to accept `role`. I am not sure if that makes sense.

mutebg (approved):

src/structured-list/__tests__/__snapshots__/structured-list.test.tsx.snap
Vanals: I guess you will have one less <li> element, as the <hr> would not need wrapping. But yeah, get it, not straight forward 👍

Issue #782, opened 2023-03-27
upgrade to webpack 5 & improve pipeline speed


jps:
# Description

- cypress tests are now taking > 10 mins to run. I looked at doing this through the CircleCI orb but cypress has been a bit scummy and made it so you can only use the feature if you sign up for their cloud service. 
- cypress-parallel seems to be the cheap and easy way to go
- Bonus webpack upgrade included as I couldn't run the e2e tests locally likely because I am using WSL, this should improve build speeds a bit

PR #650, opened 2023-02-17
build(650): upgrade to webpack 5 & improve pipeline speed

jps:
#650

**What**

1. Background - why this is needed our cypress tests are now taking > 10 mins to run. I looked at doing this through the CircleCI orb but cypress has been a bit scummy and made it so you can only use the feature if you sign up for their cloud service. 
2. What did you do cypress-parallel seems to be the cheap and easy way to go
3. Bonus webpack upgrade included as I couldn't run the e2e tests locally likely because I am using WSL, this should improve build speeds a bit
4. What should the reviewers expect? faster tests and builds

**I have done:**
- [ ] Written unit tests against changes
- [ ] Written functional tests against the component and/or NewsKit site
- [ ] Updated relevant documentation

**I have tested manually:**
- [ ] The feature's functionality is working as expected on Chrome, Firefox, Safari and Edge
- [ ] The screen reader reads and flows through the elements as expected.
- [ ] There are no new errors in the browser console coming from this PR.
- [ ] When visual test is not added, it renders correctly on different browsers and mobile viewports (Safari, Firefox, small mobile viewport, tablet)
- [ ] The Playground feature is working as expected

**Before:**

**After:**

**Who should review this PR:**

**How to test:**

PR #800, opened 2023-04-03
Build/782 parallel cypress webpack 5

jps:
#782 

**What**

- Background - why this is needed our cypress tests are now taking > 10 mins to run. I looked at doing this through the CircleCI orb but cypress has been a bit scummy and made it so you can only use the feature if you sign up for their cloud service.
- What did you do - cypress-parallel came out as the best way for us to proceed at this point in time ^1
- Bonus WebPack upgrade included as I couldn't run the e2e tests locally likely because I am using WSL, this should improve build speeds a bit and does open the way for rspack which @JohnTParsons has been looking at. 
- What should the reviewers expect? faster tests and builds

image

The following image shows some of the tunings that I looked at as part of this work. unfortunately, cypress-parallel becomes flakey at higher worker counts. It's not immediately clear what the issue is here. 

image

^1 because of the limitation / bug in cypress-parallel it may be beneficial to take a look at sorry-cypress again at some point in the future so we can better utilize large instance sizes in CirlceIC and increased benefit. Setting this up will be less simple than cypress-parallel as it requires it's orchestration service to be running. I do have an example of running this locally available in this commit https://github.com/newscorp-ghfb/NewsKit/pull/650/commits/72cd996a7846117a14b97612d5633da75aa0e0d2 

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


pp-serviceaccount:
  Documentation Site: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/782-parallel-cypress-webpack-5
  Storybook: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/782-parallel-cypress-webpack-5/storybook

cypress/site/functional/accessibility.common.js
```diff
@@ -1,19 +1,14 @@
...
+export function runA11yTestForRoute(route) {
+  describe(`Page accessibility for ${route} section`, () => {
     it(`should pass basic a11y test for ${Cypress.config(
       'baseUrl',
-    )}${path}`, () => {
-      cy.mockConsentAndVisit(path);
-      cy.wait(1000);
+    )}${route}`, () => {
+      cy.mockConsentAndVisit(route);
+      cy.wait(1000); // FIX: This accounts for 2mins of the test run time
```
JohnTParsons, line 9: I presume it fails without that wait? Perhaps it could wait for something to appear in the DOM instead?

JohnTParsons (approved):

mutebg (approved):

cypress/site/functional/accessibility.common.js
jps: I would expect it to get increasingly flakey the lower the number gets, It was in there from before, I'll spin it out into a separate task to take a look at as would like to get this merged. The Impact of this will be a bit less than 2mins given it's now parallelized. so possibly not as high priority as say compared to the rspack work.

jps:
> I've noticed that package.json contains "@types/webpack": "^4.4.34" This could be upgraded to version 5.28.1 (in this PR or a future one)

addressed in https://github.com/newscorp-ghfb/NewsKit/pull/804

PR #838, opened 2023-04-18
feat(751): add optimisation to with-own-theme

### LINKED ISSUE #751: Add optimisation to with-own-theme function
# Description
In the spike investigation about theme switching performance, we found that `with-own-theme` function takes too much time, in order to improve that we can implement a cache like in the proposed PR https://github.com/newscorp-ghfb/NewsKit/pull/706

This cache should be optional, we can add it as a `feature flag` the same way we do for CSS variables, in that way we can test on our web-site for the next-couple of months if that works correctly.

# Tasks/Steps
- copy the improved version of with-own-theme from https://github.com/newscorp-ghfb/NewsKit/pull/706
- add a feature flag `useThemeCache` ( or come up with better name ) and add the flag to theme-provider, and as a condition inside with-own-theme
- enable the flag to newskit website 

# Acceptance Criteria
*All acceptance criteria must be met before task can be marked as 'done'*

# Supporting Information
Will release without any supporting documentation for now

# Dependencies
*Any known business dependencies or reasons to engage with another team and input into their roadmap/timings*

JohnTParsons:
closes #751

**What**

Add optimisation to with-own-theme to speed up re-rendering of components (including Drawer).

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/test/test-framework-setup.ts
+++ b/src/test/test-framework-setup.ts
@@ -8,8 +8,6 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 failOnConsole({
-  // eg Warning, "Theme token "stylePresets.audioPlayerVolumeControlPopover" was not found when compiling theme!"
-  shouldFailOnWarn: false,
   silenceMessage: errorMessage => {
     if (
       errorMessage.includes(
--- a/src/test/test-framework-setup.ts
+++ b/src/test/test-framework-setup.ts
@@ -8,6 +8,8 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 failOnConsole({
+  // eg Warning, "Theme token "stylePresets.audioPlayerVolumeControlPopover" was not found when compiling theme!"
+  shouldFailOnWarn: false,
   silenceMessage: errorMessage => {
     if (
       errorMessage.includes(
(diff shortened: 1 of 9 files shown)


pp-serviceaccount:
  Documentation Site: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/751-with-own-theme
  Storybook: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/751-with-own-theme/storybook

src/test/test-framework-setup.ts
```diff
@@ -8,6 +8,8 @@ global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
 
 failOnConsole({
+  // eg Warning, "Theme token "stylePresets.audioPlayerVolumeControlPopover" was not found when compiling theme!"
+  shouldFailOnWarn: false,
```
mutebg, line 12: in the past, we decided to fail on warnings so that we don't end up with many warnings. 
That error means something with theme compiling is not correct in `src/audio-player-composable/components/volume-control/style-presets.ts`

I actually wonder if this `__extends` functionality is not broken from v3/4 when we introduced with-own-theme

JohnTParsons: I have now reverted the above change and everything runs ok, without warnings or errors.

It was only needed by the previous attempt, which used recurseUnknown() instead of compileTheme(). Since switching to compileTheme, there have been no stylePresets or extends errors or changes to the drawer snapshot file.

mutebg (approved):

jannuk59 (approved):

jps (approved):

PR #887, opened 2023-05-11
feat(851): investigate rem support

### LINKED ISSUE #851: Investigate REM support
# Purpose of the spike

Previously there were technical limitations related to font cropping that prevented us from moving to REM however since the solution was reworked and migrated to font metrics these are no more 🥇 .

At present the base NewsKit theme uses px-based sizing for fonts. Many teams across News (The Sun, The Times) use REM-based values for their fonts. There is no limitation in them moving their own themes to use REM but... there is no guarantee that all NewsKit components will render as expected as their layout will potentially be adjusted by different font sizes.

Also it's worth calling out that just because teams are using REM it doesn't mean that they are doing it correctly.

Image

Image

image

However, some sites appear to do a good job up to a certain point, this example of the BBC is at 32px, it breaks beyond this. 

image

The purpose of this spike is to conduct an audit and assess what changes will need to be made to move to REM in the base theme. 

- Update the base font sizes to be REM-based values, and look at each in the theme checker at different user-defined font sizes (24px,32px) to look for abnormalities. We suspect things like TextField, Button, Button could all cause issues due to the fact they will be different sizes
- Do we want to set any other values to be REM or EM based. 
- Do we want to consider moving to REM on the site first, the site looked quite broken at larger user-defined font sizes, what work will be required to rectify this?

# Outcomes of the Spike

- Feedback learnings to the team 
- An idea of the amount of work required to do this

# Time box
4 days

# Supporting Information

This article is worth a read https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/

Geri's Pixels vs REMs research: https://nidigitalsolutions.jira.com/wiki/spaces/NPP/pages/4380098683/PX+vs+REMS

JohnTParsons:
closes #851

**What**

For this spike, the comments on https://github.com/newscorp-ghfb/NewsKit/issues/851 are more important than the code, which is just a means to the end of getting the docs site and Storybook displaying in REMs.

What needs changing to make Storybook components look better? Not sure. All the actual components look okay to me but could do with more eyes on it to check for breaking component layouts.

SO PLEASE ADD YOUR COMMENTS to the ticket too, with suggestions for tickets required on the back of this spike, in order to finish off REM support in newskit.

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**

### DIFF
--- a/src/theme/newskit-light.ts
+++ b/src/theme/newskit-light.ts
@@ -2,4 +2,5 @@ import {createTheme} from './creator';
 
 export const newskitLightTheme = createTheme({
   name: 'newskit-light',
+  useRem: true,
 });
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1075,8 +1075,8 @@ workflows:
   pull_request:
     when:
       matches:
-        # Supporting Pr branches starting with feat/, docs/, fix/, chore/, style/, refactor/, perf/, test/, etc.
-        pattern: '^build\/.*|chore\/.*|ci\/.*|docs\/.*|feat\/.*|fix\/.*|improvement\/.*|perf\/.*|refactor\/.*|revert.*\/.*|style\/.*|test\/.*|^dependabot\/.*|^snyk-upgrade.*'
+        # Supporting Pr branches starting with feat/, docs/, fix/, chore/, style/, refactor/, perf/, test/, spike, etc.
+        pattern: '^build\/.*|chore\/.*|ci\/.*|docs\/.*|feat\/.*|fix\/.*|improvement\/.*|perf\/.*|refactor\/.*|revert.*\/.*|style\/.*|test\/.*|spike\/.*|^dependabot\/.*|^snyk-upgrade.*'
         value: << pipeline.git.branch >>
     jobs:
       - install_deps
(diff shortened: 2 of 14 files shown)


pp-serviceaccount:
  Documentation Site: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/851-rem-spike1
  Storybook: http://ncu-newskit-docs-pr.s3-website-eu-west-1.amazonaws.com/851-rem-spike1/storybook

src/theme/newskit-light.ts
```diff
@@ -2,4 +2,5 @@ import {createTheme} from './creator';
 
 export const newskitLightTheme = createTheme({
   name: 'newskit-light',
+  useRem: true,
```
mutebg, line 5: isn't it easy just to override the theme with values that are rem-based, something like this, in this case you will not need any transformation ( of course, on later stage the theme needs to be exported in rem values from figma )
```
export const newskitLightTheme = createTheme({
  name: 'newskit-light',
  overrides: {
     fonts: {
        fontSize010: "0.75rem",
        fontSize020: "0.875rem",
        fontSize030: "1rem",
        fontSize040: "1.125rem",
        fontSize050: "1.25rem",
       ...
  }
});
```

JohnTParsons: Yes, that is an easier to do it (see my comments on the ticket). For this spike, I just chose to demonstrate the harder, dynamic way (in case useful).

PR #892, opened 2023-05-12
spike(891): add createFontFacesForTheme

LukeFinch:
closes #000

**What**

2. What did you do

**I have done:**

**I have tested manually:**

**Before:**

**After:**

**How to test:**


src/theme/creator.ts
```diff
@@ -126,3 +126,37 @@ export const createTheme = ({
     {name, themeVersion: 1},
   );
 };
+
+type FontPaths = {
+  [key: string]: string;
+};
```
mutebg, line 132: Everything looks good, in an ideal world we can export this directly from the publisher and `createFontFacesForTheme` to be part of the `newskit-theme/title` so that we can add the necessary types.

like type
```
type FontPaths = {
  'Poppins': string,
  'DM Mono': string,
};
```
that way we can guarantee the people provide all fonts.

PR #929, opened 2024-03-07
(bugfix): AudioPlayerComposable - Prevent loop in firefox

### LINKED ISSUE #921: Streaming loop error in AudioPlayerComposable
# Describe the bug
When you play a live stream in the <AudioPlayerComposable /> component in Firefox (Edge, Chrome & Safari are all ok). The stream seems to loop with the first second or two segment of the stream.

# To Reproduce
1. Open Firefox Browser
2. Go to https://storybook.newskit.co.uk/?path=/docs/components-audio-player-composable--story-audio-player 
3. Press play on either the *Audio Player - full live* or *Audio Player - inline live* players
4. You can hear the loop

# Expected behavior
The stream should play normally and not get stuck on a segment

# Desktop (please complete the following information):
 - OS: Apple or Window
 - Firefox
 - 117

Note: This is using Newskit 7

lukejsikuade:
closes #921

Currently the AudioPlayerComposable component has a problem with live streaming, where it infinitely loops the first second of audio if rendered in the Firefox browser. The component seems to work fine in Chrome. The reason it doesn't work in Firefox comes down to how the two browsers handle the onDurationChange event, or more specifically the value that is applied to `duration` for unbounded streams.

In Chrome the browser assigns `duration` the value of `Infinity`, which is in accordance with the [HTML spec](https://html.spec.whatwg.org/multipage/media.html#dom-media-duration-dev) (see also [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/duration)), but in Firefox (for reasons I haven't been able to discover) the duration property is a continuously updated float.

Since we tie this continuous update to our component's `duration` state and we have that same `duration` listed as a dependency in a `useEffect`, this means that the callback for the `useEffect` is being continuously updated for unbounded streams in Firefox. That same callback is used to initialise the component by setting the current time to 0.

```
  useEffect(() => {
    if (audioRef && audioRef.current && duration) {
      audioRef.current.currentTime = initialTime;
    }
    setCurrentTime(0);
    setDisplayDuration(0);
  }, [src, initialTime, duration]);
```
This results in the infinite loop we are seeing where the audio's current time is constantly being reset as new audio data arrives and the value of `duration` changes.

Since we only really appear to want to call this function initially or when the src of the audio changes, I am proposing that we resolve the issue by removing `duration` as a factor of the component's initial setup entirely.

⚠️  There is an additional issue for Firefox that this does not solve, which is that autoplay still doesn't work.

**I have done:**

**I have tested manually:**

### DIFF
--- a/src/audio-player-composable/audio-player-composable.tsx
+++ b/src/audio-player-composable/audio-player-composable.tsx
@@ -82,7 +82,7 @@ export const AudioPlayerComposable = ({
     }
     setCurrentTime(0);
     setDisplayDuration(0);
-  }, [src, initialTime, duration]);
+  }, []);
 
   const {
     audioEvents,
--- a/src/audio-player-composable/audio-player-composable.tsx
+++ b/src/audio-player-composable/audio-player-composable.tsx
@@ -74,15 +74,19 @@ export const AudioPlayerComposable = ({
   });
 
   useEffect(() => {
+    setCurrentTime(0);
+    setDisplayDuration(0);
+  }, [src]);
+
+  useEffect(() => {
     // On render onTimeUpdate will be fired and initialTime will be set as a value for currentTime state.
     // I can't set this one to the setCurrentTime state directly as the audioElement time
     // will still be 0, currentTime will be overridden to 0 and the audio will start from 0
+
     if (audioRef && audioRef.current && duration) {
       audioRef.current.currentTime = initialTime;
     }
-    setCurrentTime(0);
-    setDisplayDuration(0);
-  }, []);
+  }, [duration, initialTime]);
 
   const {
     audioEvents,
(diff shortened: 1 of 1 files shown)


src/audio-player-composable/audio-player-composable.tsx
```diff
@@ -73,16 +73,20 @@ export const AudioPlayerComposable = ({
     currentTimeRef.current = currentTime;
   });
 
+  useEffect(() => {
+    setCurrentTime(0);
+    setDisplayDuration(0);
+  }, [src]);
```
lukejsikuade, line 79: This is similar to what's found in audio-player:
```
 useEffect(() => {
    setTrackPosition([0]);
    setDisplayDuration(0);
  }, [src]);

```

lhaggar (approved):
@ayush04 @mkurshumov do you guys use the composable player in UWP? If so then probably wise to give this a quick test in context 👍

mkurshumov:
> @ayush04 @mkurshumov do you guys use the composable player in UWP? If so then probably wise to give this a quick test in context 👍

We are not using `AudioPlayerComposable`. We are using `AudioPlayer`.
image

